PR #27, opened 2016-09-04
Offline Sprites

TheBrokenRail:
This allows scratch cat to be loaded offline IF put in the correct foilder


TheBrokenRail:
Ready To Merge?

TheBrokenRail:
@cwillisf Ready To Merge? (Along With The Scratch-VM Pull Request)

tmickel:
@cwillisf I'm going to merge this as I think it looks reasonable. Most of the changes are coming from a recompile. We definitely need to handle offline assets in a much more systematic way, but this should get people started using the renderer without internet access.

PR #38, opened 2016-09-20
Build module to be imported

rschamp:
If another package wants to import `src/index.js`, it needs to replicate the loader config in `webpack.config.js`. To avoid this, package the module in `dist.js` and set that as the module to require.

I changed `main` to point to this so that other packages can just `require('scratch-blocks')`, but it would also work for our purposes to add `"webpack": "./dist.js"` if we want to keep `src/index.js` as `main`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,26 +9,27 @@
     "type": "git",
     "url": "git+ssh://git@github.com/LLK/scratch-render-webgl.git"
   },
-  "main": "./src/index.js",
+  "main": "./dist.js",
   "scripts": {
+    "postinstall": "webpack",
     "test": "make test"
   },
   "dependencies": {
-    "svg-to-image": "1.1.3",
-    "twgl.js": "1.5.2",
-    "xhr": "2.2.0"
-  },
-  "devDependencies": {
     "babel-core": "6.9.1",
     "babel-eslint": "6.0.4",
     "babel-loader": "6.2.4",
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -42,3 +34,25 @@ module.exports = {
...
+        filename: '[name].js'
+    },
+}),
+Object.assign({}, base, {
+    entry: {
+        'render': './src/index.js'
+    },
+    output: {
+        library: 'ScratchRender',
+        libraryTarget: 'commonjs2',
```
rschamp: Apparently you can give webpack an array of configurations and it will build multiple targets. Since the output of dist.js isn't a web bundle, I couldn't just add it as an entry to the existing config and had to do this.

package.json
```diff
@@ -9,26 +9,27 @@
     "type": "git",
     "url": "git+ssh://git@github.com/LLK/scratch-render-webgl.git"
   },
-  "main": "./src/index.js",
+  "main": "./dist.js",
   "scripts": {
+    "postinstall": "webpack",
```
cwillisf: Ah, I like the combination of this and the `.gitignore` line

rschamp:
I think it's useful to supply a bundle that can be used without webpack (for prototyping etc.).  I imagine once this is on npm, we can stop having the artifacts checked in the the repo and build everything in npm prepublish.

PR #40, opened 2016-09-24
Publish to NPM and Github Pages

rschamp:
To make the `npm install` process as close to what we currently have, and to reduce version churn during development, every push releases a new version, 0.1.0-[git sha].

Using [travis-after-all](https://github.com/alrra/travis-after-all) publish after all the tests in all environments pass.

I deleted and ignored the checked-in build artifacts. When `npm install` runs, npm will run `prepublish` which causes a build, adding them back.

To publish to gh-pages, there is an environment variable `GH_TOKEN`, which stores a Personal Access Token I generated.  If this token stops working, it can be updated in the Settings panel for the repo in Travis.


rschamp:
For consistency, I just updated the way the playground is served by the dev server to match scratch-vm. No redirect from index -> playground, the playground is just the content base for the dev server (i.e., after `npm start`, go to http://localhost:8080/ for the playground).

rschamp:
@cwillisf Yes the empty `.npmignore` says to npm publish, "Don't ignore the stuff in .gitignore, just ignore the normal things you would"

package.json
```diff
@@ -4,32 +4,35 @@
...
   },
   "main": "./dist.js",
   "scripts": {
-    "postinstall": "./node_modules/.bin/webpack",
+    "prepublish": "make build",
+    "start": "make serve",
     "test": "make test"
   },
-  "dependencies": {
+  "dependencies": {},
```
thisandagain: Nit: remove this.

.travis.yml
```diff
@@ -1,8 +1,32 @@
...
+    $exitCode = 0 &&
+    # Only release on release branches
+    $RELEASE_BRANCHES =~ $TRAVIS_BRANCH &&
+    # Don't release on PR builds
+    $TRAVIS_PULL_REQUEST = "false"
+  ]]; then
+    # Authenticate NPM
+    echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
+    # Set version to commit SHA
+    npm --no-git-tag-version version $(node -p -e "require('./package.json').version")-${TRAVIS_COMMIT:0:5}
```
thisandagain: Nit: The `0.1.0-[git sha]` convention breaks with semver. Semver supports `0.1.0-[unix epoch]` or other alphanumeric increment however. An example use case would be specifying `"scratch-render": "^0.1.0"` in a `package.json` file. Using the git hash approach here could lead to unexpected results.

rschamp: I tried using the `0.1.0-[epoch]` style but it didn't seem to work with a package requiring `"scratch-render": "^0.1.0"` (npm complained that the version couldn't be found). However, I found that using `0.1.0-[release label].[increment]` does work, so I went with that.

PR #41, opened 2016-10-06
Add `setDrawableOrder` for layering

tmickel:
Here's one way to deal with layering. I tried to capture all the possible options we might want - moving drawables relative to each other, or to an absolute position in the list, and ensuring drawbles stay in front of a certain index (for the stage).

Is there anything else we might want here?

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -142,6 +142,40 @@ RenderWebGL.prototype.destroyDrawable = function (drawableID) {
 };
 
 /**
+ * Set a drawable's order in the drawable list (effectively, z/layer).
+ * Can be used to move drawables to absolute positions in the list,
+ * or relative to their current positions.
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -142,6 +142,40 @@ RenderWebGL.prototype.destroyDrawable = function (drawableID) {
...
+        // Remove drawable from the list.
+        var drawable = this._drawables.splice(oldIndex, 1)[0];
+        // Determine new index.
+        var newIndex = order;
+        if (opt_isRelative) {
+            newIndex += oldIndex;
+        }
+        if (opt_min) {
+            newIndex = Math.min(newIndex, opt_min);
+        }
```
cwillisf: I'd suggest adding here something like:
```js
newIndex = Math.max(newIndex, 0);
```
This will prevent `splice` from wrapping negative `newIndex` around to the end of the array.

tmickel: Sounds good ,fixed!

PR #42, opened 2016-10-11
Precise bounding boxes

tmickel:
Not sure if this is the right solution, but I got curious to experiment with bounding boxes and thought it might be at least worth showing to get thoughts going.

Demo: https://www.dropbox.com/s/c68l3q04qu5398f/bounds.mov?dl=0

The basic approach here is something like this: https://en.wikipedia.org/wiki/Minimum_bounding_box#Arbitrarily_oriented_minimum_bounding_box

To support things like this: https://scratch.mit.edu/projects/124995042/ correctly and efficiently.

I wanted to be able to recalculate the precise bounds without doing a read-back every time the direction/scale changes. My first thought was to calculate the pixel-precise bounding box in a single orientation, like this:
screen shot 2016-10-11 at 2 58 45 pm
And then simply transform the four corner points when the sprite is transformed. But, I think this is not enough. For example, for the cat in the orientation above, the whiskers sets the minimum bounding box. But once you rotate to something like this:

screen shot 2016-10-11 at 2 59 53 pm

The whiskers no longer matter - it's the foot instead.

This implementation works by calculating the convex hull points for the drawable whenever necessary (e.g., whenever the shape of the drawable could possibly change, including on some graphic effects). Whenever the bounds are requested, the convex hull points are transformed with the drawable transform, and then the box can be found from that.

Thoughts very welcome on both approach and details. Might be missing a better approach here, would be curious to hear!

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -3,11 +3,29 @@
 <head>
     <meta charset="UTF-8">
     <title>Scratch WebGL rendering demo</title>
+    <style>
+        #scratch-stage { width: 480px; }
+        #debug-canvas { width: 480px; }
+    </style>
 </head>
 <body style="background: lightsteelblue">
 <canvas id="debug-canvas" width="10" height="10" style="border:3px dashed red"></canvas>
 <canvas id="scratch-stage" width="10" height="10" style="border:3px dashed black"></canvas>
 <p>
+    <select id="fudgeproperty" onchange="onFudgePropertyChanged(this.value)">
+        <option value="posx">Position X</option>
+        <option value="posy">Position Y</option>
+        <option value="direction">Direction</option>
+        <option value="scalex">Scale X</option>
+        <option value="scaley">Scale Y</option>
+        <option value="color">Color</option>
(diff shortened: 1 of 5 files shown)


playground/index.html
```diff
@@ -3,11 +3,29 @@
...
+    <style>
+        #scratch-stage { width: 480px; }
+        #debug-canvas { width: 480px; }
+    </style>
 </head>
 <body style="background: lightsteelblue">
 <canvas id="debug-canvas" width="10" height="10" style="border:3px dashed red"></canvas>
 <canvas id="scratch-stage" width="10" height="10" style="border:3px dashed black"></canvas>
 <p>
+    <select id="fudgeproperty" onchange="onFudgePropertyChanged(this.value)">
```
tmickel: I also added this to the playground to let us play with multiple properties at once more easily....

tmickel:
Oh, cool - I see it now! On my first search I assumed it was something Flash was doing for us. Nice to know that this is the current approach too!

PR #43, opened 2016-10-11
Add `RenderWebGL.prototype.isTouchingDrawables`

tmickel:
Can be used to implement "sprite touching sprite."

This is mostly a copy-paste from color touching color. Specific adjustments:

- Accept a list of candidateIDs to check against, and only draw those (color touching color draws everything).
- Clear to `noneColor`, not background color.
- Remove mask specific to colors, always use silhouette mode.
- Update check to pass anytime `pixels` is populated.


tmickel:
Agreed!

PR #45, opened 2016-10-11
Expose skin/costume size

tmickel:
With an eye toward https://github.com/LLK/scratch-vm/issues/255 which I think requires it.


tmickel:
That's a good point - I think I'd actually lean towards making a copy myself...after a quick search, this might be the only place where such an object is returned? I think all are other methods are mostly returning booleans... Anyway for this I'll switch to a copy.

PR #48, opened 2016-10-12
Canvas skins for Drawables

tmickel:
Mostly intended as a demo...

I was thinking about how to implement monitors, and this could be one approach. With this, I think we even have the choice whether the actual UI is implemented in the renderer (perhaps, for example, as a subclass of Drawable), or in the VM (i.e., we can construct a canvas over there and just pass it as a skin whenever it's updated).

This was really easy to pull off, since canvas nodes are basically treated as bitmaps.

canvas-draw

Thoughts?

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -9,7 +9,6 @@
     </style>
 </head>
 <body style="background: lightsteelblue">
-<canvas id="debug-canvas" width="10" height="10" style="border:3px dashed red"></canvas>
 <canvas id="scratch-stage" width="10" height="10" style="border:3px dashed black"></canvas>
 <p>
     <select id="fudgeproperty" onchange="onFudgePropertyChanged(this.value)">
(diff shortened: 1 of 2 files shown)


tmickel:
also fun:
oct-12-2016 19-56-07

cwillisf:
This is really cool as a tech demo, but I'm not sure that the API "split" is in the right place. Specifically, I don't think we should pass something like an HTML element across the renderer's API. I'm thinking about the future where we may have a different renderer that isn't implemented in terms of WebGL and/or may have a restricted communication link (maybe JSON-like) between the renderer and its client. These could both be true for a native renderer on a mobile platform, for example. Trying to handle a `Canvas` element in that situation could be difficult to impossible.

All that said, I don't necessarily think it's a bad idea to render using canvas-based `Drawable` or `Skin` items. Maybe it would make sense to have a call that builds a `Monitor`, for example, which is a special kind of `Drawable` that can be updated with the necessary contents. Or we could have a call to create some sort of dynamic vector skin, where there might be a Canvas-like API to fill it.

tmickel:
Makes sense. Will go down one of those paths instead.

PR #49, opened 2016-10-13
Implement bubble rendering

dekrain:


tmickel:
Hi @dekrain:

I checked out your code and ran `renderer.renderBubble(1, "test");` and nothing happened. Please advise?

Please also note that the say bubble is slated for a redesign, so any implementation is subject to being rewritten in the future...

src/RenderWebGL.js
```diff
@@ -621,6 +621,60 @@ RenderWebGL.prototype._drawThese = function(
...
 
 /**
+ * Draw say bubble next to the Drawable
+ * @param {int} drawableID A Drawable ID for draw bubble to
+ * @param {string} text A text for render
+ */
+
+RenderWebGL.prototype.renderBubble = function (drawableID, text) {
+    var  drawable = Drawable.getDrawableByID(drawableID)
+        ,bounds = drawable.getBounds()
```
tmickel: Before you call `getBounds`, you need to ensure that the convex hull of the Drawable has been updated. Maybe it would be preferable to call `this.getBounds`, which already does that.

src/RenderWebGL.js
```diff
@@ -621,6 +621,60 @@ RenderWebGL.prototype._drawThese = function(
...
+        ,xy = {x: bounds.right + 5,
```
tmickel: This style of `,variable=...` doesn't exist anywhere in our codebase. Define the variable with separate `var`s instead?

src/RenderWebGL.js
```diff
@@ -621,6 +621,60 @@ RenderWebGL.prototype._drawThese = function(
...
+               y: bounds.top}; // A point for bubble render
```
tmickel: What about keeping the bubble within the bounds of the stage?

src/RenderWebGL.js
```diff
@@ -621,6 +621,60 @@ RenderWebGL.prototype._drawThese = function(
...
+
+    var canvas = typeof document !== 'undefined' ?
+                 document.createElement('canvas') : null,
+        ctx;
+    if (!canvas) return; //throw new Error('Cannot create canvas!');
+    ctx = canvas.getContext('2d');
+
+    var w = ctx.measureText(text).width,
```
tmickel: The font should be set before any `ctx.measureText`, otherwise your measurement would likely be for a different font.

src/RenderWebGL.js
```diff
@@ -621,6 +621,60 @@ RenderWebGL.prototype._drawThese = function(
...
+        h = 15, col = 0;
+
+    // Global settings
+    canvas.width = w;
+    canvas.height = 500; // Temporary
```
tmickel: Why do you need this?

src/RenderWebGL.js
```diff
@@ -621,6 +621,60 @@ RenderWebGL.prototype._drawThese = function(
...
+    ctx.fillStyle = '#000000';
+
+    // Draw text
+    ctx.font = h + 'px' + ' ' + 'sans-serif';
+    text.split('\n').forEach(function(line){
```
tmickel: Is the text not wrapped like 2.0?

tmickel:
@dekrain This doesn't work for me. Please only open pull requests that are completely functional.

liam4:
> Please only open pull requests that are functional at a basic level.

Perhaps it's works perfectly well, but the way to use it isn't as intuitive as @dekrain may have thought? :)

Anyways, giving some information on how to use it in the pull request's original post would be great..

tmickel:
@liam4 and @dekrain Yes - sorry if I'm being short! Have been feeling slightly overwhelmed with issues and pull requests across these projects.

@dekrain Appreciate the effort, if you want to re-open with some instructions of how to use and a screenshot or two of it working, that would be appreciated!

PR #51, opened 2016-10-18
Add rotation center and bitmap/costume resolution

tmickel:
This seems to work well, except for SVG backdrops for some reason. Those seem to almost always be off by a few pixels in both X and Y...it looks stage centers are a special case in 2.0: https://github.com/LLK/scratch-flash/blob/master/src/scratch/ScratchObj.as#L219 any thoughts on that?

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -60,6 +60,7 @@ class Drawable {
 
         this._position = twgl.v3.create(0, 0);
         this._scale = twgl.v3.create(100, 100);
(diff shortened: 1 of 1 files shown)


liam4:
Here's a test project I made earlier this month: https://scratch.mit.edu/projects/123898100/
..and in Scratch 3.0: http://llk.github.io/scratch-vm/#123898100

tmickel:
Hmmm, ok. Will investigate later...

Issue #29, opened 2016-09-08
Pick coordinates in Scratch-coordinate-space


tmickel:
To make queries consistent across the renderer.

PR #55, opened 2016-10-21
Bounding-box optimizations for touching color, touching drawables

tmickel:
Features:

- Adds a Rectangle utility class with lots of useful methods. The Rectangle assumes coordinates are in Scratch-space.
- Add `getAABB` to Drawable, which returns the quickly-calculated bounding box.
- Add `getFastBounds` to Drawable. If calculating the convex hull-tight bounding box would be fast (not requiring pixel-reading), it returns that; otherwise it returns the same box as `getAABB`.
- Use `getFastBounds` to optimize isTouchingColor and isTouchingDrawables. This does what was previously suggested in the TODOs: limit the size of the viewport to the query bounding box, only read the pixels from that bounding box, and filter the items drawn to possible intersecting bounding boxes. I chose to do the filtering before the `_drawThese` filter function, so that we could return very early if possible.

I didn't add the same optimizations to `pick`, because it would first help to put pick coordinates in Scratch-space (#29) to do the bounding box intersection consistently with the above.

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -462,7 +462,7 @@ Drawable.prototype.setConvexHullPoints = function (points) {
  * This function applies the transform matrix to the known convex hull,
  * and then finds the minimum box along the axes.
  * Before calling this, ensure the renderer has updated convex hull points.
- * @return {Object} Bounds for a tight box around the Drawable.
+ * @return {!Rectangle} Bounds for a tight box around the Drawable.
  */
 Drawable.prototype.getBounds = function () {
     if (this.needsConvexHullPoints()) {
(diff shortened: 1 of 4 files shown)


tmickel:
Touching color debug canvas:
color-touching-opt

src/RenderWebGL.js
```diff
@@ -243,16 +243,45 @@ RenderWebGL.prototype.getSkinSize = function (drawableID) {
...
+    if (candidateIDs.length == 0) {
+        // No possible intersections based on bounding boxes.
+        return false;
+    }
+
+    // Limit size of viewport to the bounds around the target Drawable,
+    // and create the projection matrix for the draw.
+    gl.viewport(0, 0, bounds.width, bounds.height);
+    const projection = twgl.m4.ortho(
+        bounds.left, bounds.right, bounds.bottom, bounds.top, -1, 1);
```
cwillisf: This big chunk of code is nearly identical in `isTouchingColor` and `isTouchingDrawables`... could it be pulled out into a helper function?

src/Drawable.js
```diff
@@ -488,32 +489,49 @@ Drawable.prototype.getBounds = function () {
...
+    ]);
     return bounds;
 };
 
 /**
+ * Return the best Drawable bounds possible without performing graphics queries.
+ * I.e., returns the tight bounding box when the convex hull points are already
+ * known, but otherwise return the rough AABB of the Drawable.
+ * @return {!Rectangle} Bounds for the Drawable.
+ */
```
cwillisf: Very nice 👍

src/Rectangle.js
```diff
@@ -0,0 +1,138 @@
...
+        this.left = Math.min(this.left, this.right);
+        this.right = Math.max(this.right, this.left);
+        this.bottom = Math.min(this.bottom, this.top);
+        this.top = Math.max(this.top, this.bottom);
+    }
+
+    /**
+     * Push out the Rectangle to integer bounds.
+     */
+    ceil() {
```
cwillisf: Minor nitpick:
The name `ceil`, to me, doesn't correspond well with what this function means. I think I would expect `bounds.ceil` to `ceil` each component rather than expand the bounds. I can see the argument that expanding the bounds could "feel like" a `ceil` operation, though... is that what you were going for?
Anyway, how would you feel about calling this something else? `snap` or `snapToInt` come to my mind, but if you have other thoughts I'd love to hear them. I'm also fine with leaving it `ceil` if that one feels right to you.

tmickel: `snapToInt` sounds good to me!

src/RenderWebGL.js
tmickel: I tried to refactor this into two - one to get the drawableId bounds, and one to filter the candidateIDs.

PR #56, opened 2016-10-24
A start on SVG "quirks mode"

tmickel:
I definitely don't think this is complete, but hopefully it can provide some thinking for how to approach Scratch 2.0 quirks.

Example project:
https://scratch.mit.edu/projects/127228166/

Before "quirks mode":
screen shot 2016-10-24 at 11 56 40 am
After "quirks mode":
screen shot 2016-10-24 at 11 56 45 am

Differences:
- Scratch 2.0 appears to ignore SVG width, height, and viewBox properties. Many SVGs exported from Scratch 2.0 appear to draw things in negative coordinates or otherwise outside the given viewBox, so we need to fix those in order to use the browser's SVG renderer. In this implementation, the bounds of the SVG are directly measured by asking the browser to draw it as a DOM element, and then taking `getBBox` measurements, before re-drawing the SVG again with corrected measurements. I tried to do this several different ways - this was the best way I found so far.
- Transforms text in a bunch of ways - strips out `x` and `y` properties that Scratch 2.0 appears to set incorrectly and otherwise ignore, fixes non-standard line break treatment, and adds in fonts (which was actually quite tricky).

There is no doubt in my mind that this has bugs, is incomplete (I'm sure there are more quirks), and maybe these could be done better some other way. Would be happy to have more ideas!


src/svg-quirks-mode/fonts/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
```
cwillisf: I'm not an expert here, but if I'm reading the [chart on the License Compatibility page on Wikipedia](https://en.wikipedia.org/wiki/License_compatibility#Compatibility_of_FOSS_licenses) correctly it seems we like we can't include Apache 2.0-licensed work in this repository. I'm even less sure about the OFL below.

src/svg-quirks-mode/svg-renderer.js
```diff
@@ -0,0 +1,251 @@
...
+
+        // Append the SVG dom to the document.
+        // This allows us to use `getBBox` on the page,
+        // which returns the full bounding-box of all drawn SVG
+        // elements, similar to how Scratch 2.0 did measurement.
+        let svgSpot = document.createElement('span');
+        document.body.appendChild(svgSpot);
+        svgSpot.appendChild(this._svgTag);
+        // Take the bounding box, and then destroy the element.
+        let bbox = this._svgTag.getBBox();
```
cwillisf: If there's there any chance that calling `getBBox()` will throw (either it throws internally or `this._svgTag` doesn't have a `getBBox` property) then it might make sense to put this line in a `try` with the `removeChild` call in `finally`.

cwillisf:
@tmickel do you know of a way to tell (from code) that a particular SVG was created in the Scratch 2.0 style? I'm imagining a world where some of our SVGs require this special treatment but others can be render as-is... how do we tell the difference?

tmickel:
@cwillisf Unfortunately I'm not sure if there's an easy way to accomplish what you're talking about. The Scratch 2.0 paint editor does insert a comment into SVGs, along the lines of `<!-- Scratch paint editor made this -->`. However, _all_ current SVGs are rendered in Scratch 2.0 with these quirks, whether or not they've been processed through the 2.0 paint editor. It seems possible to me that people have adjusted their externally-edited/created SVGs to deal with the quirks - by trial and error, or whatever. So, I'd recommend that we render all vector graphics from 2.0 with this.

But, it would make sense to me for _future_ assets and imports to use a different renderer (ideally a pure browser renderer that respects viewboxes and whatever, maybe). In that case we should probably explicitly mark new assets to distinguish when we can use a "standards mode" renderer? Any thoughts @thisandagain ?

src/svg-quirks-mode/fonts/LICENSE.txt
thisandagain: Agreed. We won't be able to mix APL with BSD-3-Clause. We may want to bring in the fonts as a dependency rather than checking them in.

thisandagain:
Thanks for putting this together @tmickel . This looks like a great start.

I like the idea of tagging new SVGs so we can use a less "quirky" rendering approach and (hopefully) split with some of the pain that we see in 2.0:
https://github.com/LLK/scratch-flash/issues/963
https://github.com/LLK/scratch-flash/issues/1175
https://github.com/LLK/scratch-flash/issues/1200
https://github.com/LLK/scratch-flash/issues/1103
https://github.com/LLK/scratch-flash/issues/411
https://github.com/LLK/scratch-flash/issues/135

chooper100:
How do you plan to handle the case for a SVG file being imported?

For example, an SVG file made in Inkscape or something similar would work best being rendered by the browser, but an SVG that had previously been downloaded from Scratch might work best in quirks mode.

My suggestion would be to treat all imported SVGs as normal SVGs (ie not quirks mode)

Also:
> people have adjusted their externally-edited/created SVGs to deal with the quirks - by trial and error, or whatever.

I've been using Inkscape for SVGs for quite a while now. Whilst Scratch 2.0s support for more advanced SVG features (eg. shadows) is ... interesting, I never had any quirks mode problems myself

src/svg-quirks-mode/fonts/LICENSE.txt
tmickel: @thisandagain Where should that go? I could import scratch-flash as a dependency, or do you want a new repo scratch-fonts?

tmickel:
@chooper100 Any SVG that could have previously been rendered by Scratch 2.0 would be rendered in quirks mode. That includes SVGs created by external tools, and SVGs created in the paint editor. Treating import-SVGs differently would no doubt lead to broken projects, because they are rendered in 2.0 with these quirks and possibly more. I'm not sure I understand what you're saying, I don't think these are treated differently in scratch-flash if they're made in external tools. Do you have an example of an SVG created in Inkscape, for example, where the viewBox is smaller than the graphic, and Scratch renders the clipped graphic?

cwillisf:
My interpretation of @chooper100's comments:
- When Scratch 3.0 loads a Scratch 2.0 project, the SVGs should render in "quirks mode"
- When Scratch 3.0 imports an SVG independently, it should render without "quirks mode"
Is that what you had in mind, @chooper100?

I think we can continue to discuss when (and when not) to use Scratch 2.0 quirks-mode SVG rendering independently of this PR. As long as we want to do it in at least one case, we'll need some code like this :)

src/svg-quirks-mode/fonts/LICENSE.txt
thisandagain: @tmickel Unfortunately I think a repo just dedicated to the fonts would be best.

chooper100:
@tmickel Yes, I meant what @cwillisf said

I've never had a problem where an Inkscape SVG has been clipped

tmickel:
@cwillisf Comments addressed I think - any other thoughts?

By the way, I tested in Chrome, Firefox, and Safari. I didn't test in IE or Edge.

tmickel:
Update:
- In my VM, it looks like IE 11 wasn't even rendering SVGs before this pull request. Not sure why - there are no console complaints.
- Edge was rendering SVGs before this pull request, but not after. Also no console complaints.

CosmicWebServices:
👏

PR #64, opened 2016-12-12
Update configuration to match other Scratch 3.0 projects

cwillisf:
Changes include:
- Update Webpack config with what I've learned from [LLK/scratch-storage](/LLK/scratch-storage)
- Start using [LLK/eslint-config-scratch](/LLK/eslint-config-scratch)
- Fix lint errors & warnings

There should be no changes in functionality.

NOTE: This change moves the build products generated by `npm run build`. I believe there are no changes needed in [LLK/scratch-vm](/LLK/scratch-vm) and [LLK/scratch-gui](/LLK/scratch-gui) since those just use `require('scratch-render')` which will use this module's `package.json` to find the build output in its new location. If there are clients using `dist.js`, `render.js`, or `render.min.js` directly then they will need to update their references for the new output locations & names.

The location changes are as follows:
- `dist.js` is now `dist/node/scratch-render.js`
- `render.js` is now `dist/web/scratch-render.js`
- `render.min.js` is now `dist/web/scratch-render.min.js`


src/ShaderManager.js
```diff
@@ -70,15 +74,15 @@ ShaderManager.EFFECT_INFO = {
...
             return Math.max(-100, Math.min(x, 100)) / 100;
         },
         shapeChanges: false
     },
     ghost: {
         mask: 1 << 6,
-        converter: function(x) {
-            return 1 - Math.max(0, Math.min(x, 100)) / 100;
+        converter: function (x) {
+            return 1 - (Math.max(0, Math.min(x, 100)) / 100);
```
rschamp: This is a good place for an arrow function.

liam4:
> I.e., use ES6-style class definition, but only define the constructor there. To me this seems like an odd half-use of ES6 classes, where all functions can be defined inside the class definition.

Just out of curiosity's sake, if you're going to use ES6 classes for constructors (because they make it more clear that the function is to be used as a class?), why not go the whole way and properly use it for methods as well?

Somebody *must* have had a reason to do that at some point... hopefully it's better than "we don't want it to be *too* ES6-y!" :)

rschamp:
I don't know if this is why the project was set up that way, but one reason to use prototype assignment for instance methods is that it gives you auto-binding for free.  If the methods are defined in the `class`, then you must bind them in the constructor (`this.method = this.method.bind(this);`) so that `this` is defined when called from an instance.

TheBrokenRail:
LLK/scratch-storage?

cwillisf:
@TheBrokenRail coming soon :D

PR #65, opened 2016-12-13
Convert class methods to ES6 style

cwillisf:
Also:
- Un-wrap lines which were originally written for an 80-column limit but don't need to wrap now that the limit is 120 columns.
- Get rid of `instance` variables that were being used to save `this` for callbacks now that those callbacks have been converted from `function` to `=>`.

In some ways this is a followup to #64, but these changes aren't specifically related to Webpack or lint config changes.

I highly recommend looking at this with white-space differences ignored. To do so, add "?w=1" (or "&w=1" if there's already a "?") to the end of the URL when on the "Files changed" view.

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -209,7 +209,6 @@ class Drawable {
      */
     _setSkinSVG (skinMd5ext) {
         const url = skinMd5ext;
-        const instance = this;
 
         const svgCanvas = document.createElement('canvas');
         const svgRenderer = new SvgRenderer(svgCanvas);
@@ -217,7 +216,7 @@ class Drawable {
         const gotSVG = (err, response, body) => {
             if (!err) {
                 svgRenderer.fromString(body, () => {
-                    instance._setSkinCore(svgCanvas, svgRenderer.getDrawRatio());
+                    this._setSkinCore(svgCanvas, svgRenderer.getDrawRatio());
                 });
             }
         };
@@ -235,10 +234,9 @@ class Drawable {
      * @private
      */
     _setSkinCore (source, costumeResolution) {
-        const instance = this;
         const callback = (err, texture, sourceInCallback) => {
-            if (!err && (instance._pendingSkin === texture)) {
-                instance._useSkin(texture, sourceInCallback.width, sourceInCallback.height, costumeResolution);
+            if (!err && (this._pendingSkin === texture)) {
+                this._useSkin(texture, sourceInCallback.width, sourceInCallback.height, costumeResolution);
             }
         };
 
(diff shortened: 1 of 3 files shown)


src/Drawable.js
```diff
@@ -78,484 +85,473 @@ class Drawable {
...
+Drawable._DEFAULT_SKIN = {
+    squirrel: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
+    '7e24c99c1b853e52f8e7f9004416fa34.png/get/',
+    bus: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
+    '66895930177178ea01d9e610917f8acf.png/get/',
+    scratch_cat: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
+    '09dc888b0b7df19f70d81588ae73420e.svg/get/',
+    gradient: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
+    'a49ff276b9b8f997a1ae163992c2c145.png/get/'
+}.squirrel;
```
rschamp: In the newest eslint-config-scratch, lines with URLs are exempt from `max-len`.

src/Drawable.js
```diff
@@ -78,484 +85,473 @@ class Drawable {
...
+     * the color are zero, the result will be Drawable.NONE; otherwise the result
+     * will be a valid ID.
+     * @param {int} r The red value of the color, in the range [0,255].
+     * @param {int} g The green value of the color, in the range [0,255].
+     * @param {int} b The blue value of the color, in the range [0,255].
+     * @param {int} a The alpha value of the color, in the range [0,255].
+     * @returns {int} The ID represented by that color.
+     */
+    // eslint-disable-next-line no-unused-vars
+    static color4bToID (r, g, b, a) {
```
rschamp: Why take `a`?

src/Drawable.js
```diff
@@ -515,11 +515,9 @@ class Drawable {
      * @param {int} r The red value of the color, in the range [0,255].
      * @param {int} g The green value of the color, in the range [0,255].
      * @param {int} b The blue value of the color, in the range [0,255].
-     * @param {int} a The alpha value of the color, in the range [0,255].
      * @returns {int} The ID represented by that color.
      */
-    // eslint-disable-next-line no-unused-vars
-    static color4bToID (r, g, b, a) {
+    static color4bToID (r, g, b) {
```
liam4: I don't speak Complicated Math or scratch-render, but I wonder if this should be color3btoID..?

rschamp: You can still pass it a fourth argument, it just doesn't matter what it is.

PR #67, opened 2016-12-29
Move `Skin` functionality out of `Drawable` into its own classes

cwillisf:
Moving `Skin` functionality (corresponding to costumes) out of `Drawable` (corresponding to sprites & clones) allows multiple `Drawable` instances to share the same `Skin` and allows a `Drawable` to change between `Skin`s without loading the image fresh each time.

In addition, this change moves management of `Drawable` IDs out of `Drawable` and into the renderer, which means that `Drawable` no longer has a "split personality" (static vs. instance). It also makes debugging easier: if you can access the renderer you can inspect all relevant state.

Setting a `Skin` from a URL has been improved dramatically by this change, but really the renderer shouldn't be responsible for downloading and parsing image files. The new `createBitmapSkin` and `createSVGSkin` methods should be used after the renderer's client (the VM, GUI, etc.) obtains an image through a download, the paint editor, etc.

This resolves LLK/scratch-vm#372

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -69,9 +69,8 @@ class Drawable {
      * Dispose of this Drawable. Do not use it after calling this method.
      */
     dispose () {
-        if (this._id >= 0) {
-            delete Drawable[this._id];
-        }
+        // Use the setter: disconnect events
+        this.skin = null;
     }
 
     /**
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -2,6 +2,7 @@ const twgl = require('twgl.js');
 const xhr = require('xhr');
 
 const Rectangle = require('./Rectangle');
+const RenderConstants = require('./RenderConstants');
 const SvgRenderer = require('./svg-quirks-mode/svg-renderer');
 const ShaderManager = require('./ShaderManager');
 
@@ -84,14 +85,6 @@ class Drawable {
 
         // Load a real skin
         this.setSkin(Drawable._DEFAULT_SKIN);
-    }
-
-    /**
-     * An invalid Drawable ID which can be used to signify absence, etc.
-     * @type {int}
-     */
-    static get NONE () {
-        return -1;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


cwillisf:
Hmm... I was thinking that I made it so they take a reference to the renderer for consistency with Drawable, but it appears that I never committed a version of Drawable that takes a reference to the renderer. Heh.

Taking just `gl` should be fine -- I'll change it. Circular references tend to feel a little like circular references anyway, and if YAGNI turns out false in this case we can always change it back later.

cwillisf:
Oh - I found the reason. Hmm...

My upcoming pen code includes the following changes:
- The renderer itself is an event emitter
- There's a class called `PenSkin` which handles all the stuff you need for a pen layer
- `PenSkin` listens to an event fired by the renderer, so it takes a reference to the renderer in its constructor
- In order to unregister for the event, `PenSkin` holds the reference until `Dispose()` time

Specifically, `PenSkin` listens for a stage resize event so that the pen layer can always match the size of the stage.

@rschamp thoughts? Would you prefer a separate event emitter rather than emitting from the renderer itself? Some other way to handle resize notification?

rschamp:
In the case of the renderer emitting resize events which `PenSkin` should listen to, I think we could still handle that in the renderer, by attaching the `PenSkin`'s listener from the renderer when the skin is constructed, and unregister it when it's destroyed. This is assuming the renderer constructs and destroys the skins (sorry I am not familiar enough with how it works yet).

But I think since this is dependent on the other work you're doing, we should keep it as-is and refactor later — that sort of thing won't be hard to change.

src/Drawable.js
```diff
@@ -122,140 +82,47 @@ class Drawable {
...
-        };
-        const willCallCallback = typeof source === 'string';
-        this._pendingSkin = twgl.createTexture(gl, options, willCallCallback ? callback : null);
-
-        // If we won't get a callback, start using the skin immediately.
-        // This will happen if the data is already local.
-        if (!willCallCallback) {
-            callback(null, this._pendingSkin, source);
-        }
+    getEnabledEffects () {
```
thisandagain: Should this also be a getter?

cwillisf: Yeah... there are quite a few things that could and maybe should be converted to getters in here. This particular one didn't actually change in this PR: it just looks like it changed because of the other stuff that moved around near it. I'm planning to address things like this in a post-pen cleanup.

PR #70, opened 2017-01-13
Calculate rotation center if it's not supplied

rschamp:
When a new skin is added, previously the rotation center was set to `[0, 0]`.  In the case of costumes and backdrops added from libraries, the center should be calculated from the bounding box of the imported skin.

Toward LLK/scratch-gui#18

### DIFF
--- a/src/BitmapSkin.js
+++ b/src/BitmapSkin.js
@@ -55,9 +55,10 @@ class BitmapSkin extends Skin {
      * Set the contents of this skin to a snapshot of the provided bitmap data.
      * @param {ImageData|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement} bitmapData - new contents for this skin.
      * @param {int} [costumeResolution=1] - The resolution to use for this bitmap.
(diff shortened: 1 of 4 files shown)


rschamp:
> One complication would be if the client creates a skin, then sets the rotation center, then the skin finishes downloading: we'd need to avoid overriding the user-set rotation center in that case.

I thought about always calculating the center, but it would require us to deal with the edge case of manually setting the center while the skin is still downloading (since we would start the download and then immediately set the rotation center [here](https://github.com/LLK/scratch-render/pull/70/files#diff-0e66131b35a2396bf8b007193ee7bc77R696)).

What if we passed in `rotationCenter` to `createSkinFromURL` rather than the `calculateRotationCenter` flag, and calculate it if `rotationCenter` is undefined?  It wouldn't help the case of someone changing the center before the skin downloads, but in this situation it would be an edge case we could deal with later.

cwillisf:
> What if we passed in rotationCenter to createSkinFromURL rather than the calculateRotationCenter flag, and calculate it if rotationCenter is undefined?

That sounds good to me.

Issue #6, opened 2016-07-06
Pen support


tmickel:
Thoughts:
- Provide a method to draw a line from (x1, y1) to (x2, y2) at color (RGB), width (px). VM handles color calculations and deciding when to draw lines.
- Stamp method for current drawable and clear method also provided.

I believe scratch-flash's pen draws the equivalent of canvas `lineCap`="round". In doing the HTML5 player, I also remember something about drawing lines at .5px off to prevent blur, but there might be a better way to deal with that.

BookOwl:
Don't forget that the Scratch pen supports alpha.

thisandagain:
Notes from meeting with @cwillisf 

#### Approach
- `Pen` is a `Drawable` with a specific type of skin which provides drawing methods
- `Pen` includes three public methods
    - `draw` (x1, y1, x2, y2, width, rgba)
    - `clear`
    - `stamp` (x, y, drawable_id)
- Set pen layer to be the render target and render quads directly

#### Problems
- Getting the lines to look correct
    - Render a quad 
    - Write a shader
    - `discard` around the radius to get "round" line caps
- Can you write on top of the existings buffer
    - Draw the current pen layer otherwise we'll need to "double buffer" (write self, write line, to pen layer)

#### Questions
- How are we thinking about HDPI screens?
    - Resolution of the display
    - Resolution of the query layer (touching color)

liam4:
Might be worth throwing out that you could be able to stamp pen data!

(`Pen instanceof Drawable`; `aPen.stamp(aDrawable)`; `aPen.stamp(aPen)`)

tmickel:
@thisandagain @cwillisf 

Can you explain a little more about why drawing lines this way vs. using Canvas 2D's existing line drawing methods? I was imagining that Canvas 2D would give us the same exact lines as Scratch 2.0, and we wouldn't have to deal with a line-cap shader or antialiasing since that would be handled by canvas/Skia/etc.

cwillisf:
Hmm... that could be worth investigating for lines, though I think stamp would become more difficult in that situation. Still, getting good pen edges and line caps in exchange for more stamping work might be a good trade.

Part of the reason I hadn't considered it at first is because I assumed this wasn't possible to do this, but here's at least one person claiming it works:
http://stackoverflow.com/questions/22570235/how-do-i-use-an-html5-canvas-as-a-webgl-texture
I thought I had heard that it was forbidden in order to prevent browser fingerprinting and stuff like that.

[...]

PR #72, opened 2017-01-19
Pen

cwillisf:
### Resolves

This set of changes adds rendering support for Scratch 2.0-style pen features as a step toward LLK/scratch-render#6.

### Proposed Changes

This introduces a new class, called `PenSkin`, which acts like an initially invisible "costume" that has pen-related functionality. The renderer's API can be used to create a `PenSkin` and assign it to a `Drawable` corresponding to the pen "layer" in Scratch, then call pen-related methods on the renderer to draw on the `PenSkin`.

Since the `PenSkin` is hosted by a regular `Drawable`, in theory we could implement some features that Scratch 2.0 did not have. This includes multiple pen layers, graphical effects on pen layers, moving pen layers, and more. This might not be desirable in core Scratch 3.0, but I think we should consider allowing extensions access to these features.

### Reason for Changes

Implementing pen features brings us closer to Scratch 2.0 compatibility.

### Test Coverage

The renderer does not currently have tests.


src/PenSkin.js
```diff
@@ -0,0 +1,185 @@
...
+                mag: gl.NEAREST,
+                min: gl.NEAREST,
+                wrap: gl.CLAMP_TO_EDGE,
+                src: this._canvas
+            }
+        );
+        this._canvasDirty = true;
+    }
+
+    _setAttributes (context, penAttributes) {
```
thisandagain: Missing jsdoc.

src/RenderWebGL.js
```diff
@@ -496,17 +540,15 @@ class RenderWebGL {
             gl.disable(gl.STENCIL_TEST);
         }
 
-        const pixels = new Buffer(bounds.width * bounds.height * 4);
+        const pixels = new Uint8Array(bounds.width * bounds.height * 4);
```
thisandagain: Do you know what babel will convert this to? `Uint8Array` is currently only supported by Chrome and Firefox. Perhaps something we can polyfill?

griffpatch:
I asked back in the scratch 2 beta for a way to have the pen as the topmost
layer... Oh how useful that would have been! :)

PR #80, opened 2017-02-07
Fix SVG size calculation

rschamp:
It was being incorrectly calculated based on devicePixelRatio.

Fixes LLK/scratch-gui#73. Thanks to @cwillisf for figuring it out!

### DIFF
--- a/src/svg-quirks-mode/svg-renderer.js
+++ b/src/svg-quirks-mode/svg-renderer.js
@@ -73,6 +73,13 @@ class SvgRenderer {
     }
 
     /**
+     * @return {[number,number]} the natural size, in Scratch units, of this SVG.
+     */
+    get size () {
+        return [this._measurements.width, this._measurements.height];
+    }
+
+    /**
      * Transforms an SVG's text elements for Scratch 2.0 quirks.
      * These quirks include:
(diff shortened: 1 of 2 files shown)


src/svg-quirks-mode/svg-renderer.js
```diff
@@ -73,6 +73,13 @@ class SvgRenderer {
     }
 
     /**
+     * @return {[number,number]} the natural size, in Scratch units, of this SVG.
+     */
+    get size () {
```
rschamp: Open to suggestions for this name. It seems odd for a renderer to have a size, but I couldn't think of anything better to call it.

liam4: The only thing that seems particularly obvious to me is `scratchSize` – since it's in "Scratch units", would that be better?

cwillisf: I'd be OK with `size`, `scratchSize`, `measurements`, or maybe `logicalSize` (a generic version of `scratchSize`). I think I'd prefer sticking with `size` for brevity and let the comment explain why it's there & what it means.

thisandagain:
It works! ✨

PR #82, opened 2017-02-09
Feature fencing

griffpatch:
Provide a method in the scratch-render to return the fenced coordinates of a drawable moving to a given position.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -739,7 +739,8 @@ class RenderWebGL extends EventEmitter {
     /**
      * Update the position object's x & y members to keep the drawable fenced in view.
      * @param {int} drawableID - The ID of the Drawable to update.
-     * @param {object} position - An object with x & y members to be fenced - These values are updated in the object to become fenced
+     * @param {object} position - An object with x & y members to be fenced
+     *                          - These values are updated in the object to become fenced
      */
     getFencedPositionOfDrawable (drawableID, position) {
         const drawable = this._allDrawables[drawableID];
@@ -753,7 +754,8 @@ class RenderWebGL extends EventEmitter {
 
         const aabb = drawable.getFastBounds();
 
-        // This is my best guess at the fencing in Scratch 2, but I suspect it may need further work to be precisely the same?
+        // This is my best guess at the fencing in Scratch 2,
+        // but I suspect it may need further work to be precisely the same?
         const sx = this._xRight - Math.min(15, Math.floor((aabb.right - aabb.left) / 2));
         if (aabb.right + dx < -sx) {
             position.x = drawable._position[0] - (sx + aabb.right);
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -737,6 +737,38 @@ class RenderWebGL extends EventEmitter {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -737,6 +737,40 @@ class RenderWebGL extends EventEmitter {
     }
 
     /**
+     * Update the position object's x & y members to keep the drawable fenced in view.
+     * @param {int} drawableID - The ID of the Drawable to update.
+     * @param {object} position - An object with x & y members to be fenced
+     *                          - These values are updated in the object to become fenced
+     */
+    getFencedPositionOfDrawable (drawableID, position) {
```
cwillisf: By our convention, a function whose name starts with `get` is usually intended to retrieve or calculate a value, often with few or no caller-visible side effects. Since this one doesn't return anything and is entirely about side effects, I think it would make sense to change the name. Maybe `setFencedPositionOfDrawable`?

src/RenderWebGL.js
```diff
@@ -737,6 +737,40 @@ class RenderWebGL extends EventEmitter {
...
+            // Right now this happens so much on some projects that a warning or exception here can hang the browser.
+            return;
+        }
+        const dx = position.x - drawable._position[0];
+        const dy = position.y - drawable._position[1];
+
+        const aabb = drawable.getFastBounds();
+
+        // This is my best guess at the fencing in Scratch 2,
+        // but I suspect it may need further work to be precisely the same?
```
cwillisf: The Scratch 2.0 implementation can be found in the function `keepOnStage` in ScratchSprite.as:
https://github.com/LLK/scratch-flash/blob/6651badebc4b4ee9d63ad7565bae27c5cb6611d0/src/scratch/ScratchSprite.as#L191

It looks to me like your implementation should have the same effect for a sprite with a non-empty costumes. In particular, 18 - 3 = 15.

If I'm reading the Scratch 2.0 code right, for a sprite using an empty costume -- where `width == height == 0` -- it effectively moves the fence outward by 3 units. I can't think of a reason for that... can you? If not, I'm happy with dropping that inconsistency until we find a reason to bring it back.

src/RenderWebGL.js
```diff
@@ -737,6 +737,40 @@ class RenderWebGL extends EventEmitter {
...
+        const sx = this._xRight - Math.min(15, Math.floor((aabb.right - aabb.left) / 2));
```
cwillisf: Could you pull the 15 out into a `const` and give it a name? I think it's fine for it to live in this function, but I think giving it a name would make this code more readable to a new Scratch contributor.

griffpatch:
I've renamed the method. I took a look at the original action script code
too. I am actually quite pleased with my observations being so accurate
(modest grin). It's a little odd how it works in Scratch 2 though as it
inflates the rectangle, and then checks an extra 15 pixels anyhow... I
think it must have evolved that way... perhaps originally the fencing was
only 3 pixels... but then it got enlarged... but it a different way?
Anyhow, I think the 0 pixel case is not inflated only because that would
cause the empty rectangle constant to get larger every time... So I guess
I'd say keep it as is for now... If it gets flagged up in any text projects
then we can change the algorithm later?

Griffpatch.

PR #90, opened 2017-02-15
SVG: compensate for viewbox offset

### LINKED ISSUE #62: viewBox and rotation center mysteries
We noticed some strange things about Carl's Cachalot project:
https://llk.github.io/scratch-vm/#73724746

Even though the trees always have y=-10, they sometimes appear very high in the sky! It appears this is the case because there is one costume that has a rotation center in some other coordinate system than the one that's expected (possibly affected by the SVG viewBox). That costume - the one that's positioned incorrectly - is notably generated by Illustrator instead of the Scratch paint editor. :(

A similar thing can be seen with the whale sprite. One of the costumes causes the whale to appear in the top right of the stage. It has a rotation center that is different from the rotation centers of the other two costumes, and that SVG was generated by Illustrator. It's hard to see the whale sprite at first but you can get it if you slow the speed of the project down using the arrow keys.

This code looks possibly suspect:
https://github.com/LLK/scratch-flash/blob/abac440f1f34932679bd001d96de2315be7020ee/src/scratch/ScratchCostume.as#L250

But I wouldn't know where to begin with this.

cwillisf:
### Resolves

Resolves #62

Sprite alignment is still not 100% perfect in comparison to Scratch 2.0, but I don't think the remaining problems are related to the SVG viewbox.

### Proposed Changes

This change adjusts the rotation center of an `SVGSkin` by an amount corresponding to the origin of the SVG's viewbox. The math wasn't as tricky as I expected ;)

### Reason for Changes

In Scratch 2.0, an SVG costume's origin is stored relative to that SVG's viewbox. In order to use the origin values correctly, then, we need to compensate for the viewbox offset. Visually, this change is necessary to make sure SVG costumes appear in the correct place whether they were authored in Scratch (which always uses a viewbox origin of [0,0]) or another tool (which may use other coordinates).

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -41,6 +41,16 @@ class SVGSkin extends Skin {
     }
 
     /**
+     * Set the origin, in object space, about which this Skin should rotate.
+     * @param {number} x - The x coordinate of the new rotation center.
+     * @param {number} y - The y coordinate of the new rotation center.
+     */
+    setRotationCenter (x, y) {
+        const viewOffset = this._svgRenderer.viewOffset;
+        super.setRotationCenter(x - viewOffset[0], y - viewOffset[1]);
+    }
+
+    /**
      * @param {[number,number]} scale - The scaling factors to be used.
      * @return {WebGLTexture} The GL texture representation of this skin when drawing at the given scale.
      */
--- a/src/svg-quirks-mode/svg-renderer.js
+++ b/src/svg-quirks-mode/svg-renderer.js
@@ -81,6 +81,13 @@ class SvgRenderer {
     }
 
     /**
+     * @return {[number,number]} the offset (upper left corner) of the SVG's view box.
+     */
+    get viewOffset () {
+        return [this._measurements.x, this._measurements.y];
+    }
+
+    /**
      * Transforms an SVG's text elements for Scratch 2.0 quirks.
      * These quirks include:
      * 1. `x` and `y` properties are removed/ignored.


cwillisf:
@rschamp here's an attempt... it's a bit hard to explain so hopefully this makes sense ;)

image

The parts of this diagram:
- the whale is the costume
- the dark box around the whale is Scratch's calculated size -- this would be the bounds if it were a bitmap, for example.
- the red marker is the rotation center. When you switch between costumes, the new costume will be shifted so that its red marker matches up with the location and orientation of the marker on the previous costume.
- the axes represent the origin of "costume space"
- the blue arrow represents the effect of this PR

Note that the relative locations of these elements aren't measured carefully -- I just eyeballed everything.

The "whale2" costume worked correctly before this change since its `viewBox` starts at 0,0 but the "whale1" costume did not. Prior to this change, the whale1 costume would appear above and to the right of where it was supposed to be. This is because Scratch 3.0 was treating the body of the costume the same way it treats bitmaps: assume that the body is between `[0,0]` and `[width,height]` -- this is depicted by the fact that the whale is near the origin in each image. The rotation center is stored in terms of the original SVG coordinate system, though, so it was being set to -115, 181 which is to the lower left. When switching costumes, the renderer would compensate for this offset by shifting the costume's contents to the upper right.

This change subtracts the origin of the viewbox (-167.5, 138.4 in this example) from the coordinates of the rotation center (-115, 181) resulting in an adjusted rotation center (52.5, 42.6). You can think of this as moving the rotation center from "SVG space" into "texture space" to parallel what `SvgRenderer`'s `_draw()` method is doing with the image itself.

rschamp:
Perfect, thanks @cwillisf!

PR #97, opened 2017-02-17
Add method to retrieve drawable pixel data

rschamp:
Towards LLK/scratch-gui#66

Add `extractDrawable`, which returns raw pixel data and other data convenient for dragging this image, for a `drawableID` picked at `x`, `y`.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -656,6 +656,78 @@ class RenderWebGL extends EventEmitter {
     }
 
     /**
+     * @typedef DrawableExtraction
+     * @property {Uint8Array} data Raw pixel data for the drawable
+     * @property {int} width Drawable bounding box width
+     * @property {int} height Drawable bounding box height
+     * @property {int} x The x coordinate relative to drawable bounding box
+     * @property {int} y The y coordinate relative to drawable bounding box
+     */
+
+    /**
+     * Return drawable pixel data and picking coordinates relative to the drawable bounds
+     * @param {int} drawableID The ID of the drawable to get pixel data for
+     * @param {int} x The client x coordinate of the picking location.
+     * @param {int} y The client y coordinate of the picking location.
+     * @return {?DrawableExtraction} Data about the picked drawable
+     */
+    extractDrawable (drawableID, x, y) {
+        const drawable = this._allDrawables[drawableID];
+        if (!drawable) return;
+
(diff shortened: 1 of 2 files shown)


src/RenderWebGL.js
```diff
@@ -452,15 +452,14 @@ class RenderWebGL extends EventEmitter {
...
-                extraUniforms);
+                {extraUniforms});
 
             gl.stencilFunc(gl.EQUAL, 1, 1);
             gl.stencilOp(gl.KEEP, gl.KEEP, gl.KEEP);
             gl.colorMask(true, true, true, true);
 
             this._drawThese(candidateIDs, ShaderManager.DRAW_MODE.default, projection,
-                testID => testID !== drawableID
+                {idFilterFunc: testID => testID !== drawableID}
```
cwillisf: I like the extra clarity on the caller side 👍

src/RenderWebGL.js
```diff
@@ -656,6 +655,79 @@ class RenderWebGL extends EventEmitter {
...
+    /**
+     * Return drawable pixel data and picking coordinates relative to the drawable bounds
+     * @param {int} drawableID The ID of the drawable to get pixel data for
+     * @param {int} x The client x coordinate of the picking location.
+     * @param {int} y The client y coordinate of the picking location.
+     * @return {?DrawableExtraction} Data about the picked drawable
+     */
+    extractDrawable (drawableID, x, y) {
+        const drawable = this._allDrawables[drawableID];
+        if (!drawable) return;
```
cwillisf: This is a weakness in my understanding of JSDoc and JS convention:
Is it more conventional to just `return` here, or to explicitly `return null`?
Is `undefined` included in `{?DrawableExtraction}`?

Airhogs777: It looks like the question mark specifically means it can either return the listed type or null: http://usejsdoc.org/tags-type.html

Airhogs777: in my opinion it's better to return null if the function is expected to return something. Return by itself is fine for void functions, but otherwise returning null proves it was intentional. Plus from what I've seen the rest of this project has usually erred on the side of avoiding using undefined when possible.

rschamp: Right, it should return `null`... this is left over from how I started writing it, by copying from the stamp method.

PR #102, opened 2017-02-20
Generate docs from src, publish to gh-pages

### LINKED ISSUE #58: Add documentation on wiki about code
- [ ] Modifying Playground
- [ ] Graphic Effects
- [ ] What JS Objects are for and Their Functions

rschamp:
This is a start toward having technical docs for the Scratch 3.0 repos.  Although it's not really used anywhere, I included the markdown plugin so that we can format our descriptions in the future.  Currently there is no link to it, but it will just publish to llk.github.io/scratch-render/docs.

Fix any broken JSDoc syntax. Unfortunately there is no way to document tuples like `{[int, int]}` for a `[width, height]` array. We could possibly fix this and make the code more readable with `{width: number, height: number}` instead of tuples?

Remove single `@fileoverview` from Rectangle.js, since it was the only file with one and was weird to have that one module show up on the docs home page.

Add a `docs` script to `package.json`, run it from `npm test`, so invalid docs will cause tests to fail (eventually we may want to start using eslint-plugin-jsdoc to catch these errors with the linter alone).

I also converted our `TODO`s to `@todo`s because why not?

Fixes #58


TheBrokenRail:
This would probably fix most of #58

src/RenderWebGL.js
```diff
@@ -164,6 +170,7 @@ class RenderWebGL extends EventEmitter {
      * @param {int} width - the new width to set.
      * @param {int} height - the new height to set.
      * @private
+     * @fires RenderWebGL#event:NativeSizeChanged
```
cwillisf: Heh - I didn't know about that one!

PR #132, opened 2017-04-11
Fix typed array exception in Edge

a49594a:
### Resolves

fix exception "Invalid offset/length when creating typed array" in ms edge browser

### Proposed Changes

just convert the float to integer use Math.floor

new Uint8Array(Math.floor(bounds.width * bounds.height * 4))

### Reason for Changes

new Uint8Array(10.1)
if the parameter of the Uint8Array constructor is not integer, chrome browser ignor the 0.1, but the edge browser throw an exception "Invalid offset/length when creating typed array".

just run the playground of scratch-vm or the scratch-gui in the edge browser, and use f12 develop tools to change the size of the stage, then you will got the exception

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -413,7 +413,7 @@ class RenderWebGL extends EventEmitter {
             gl.disable(gl.STENCIL_TEST);
         }
 
-        const pixels = new Uint8Array(bounds.width * bounds.height * 4);
+        const pixels = new Uint8Array(Math.floor(bounds.width * bounds.height * 4));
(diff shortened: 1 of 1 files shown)


a49594a:
new Uint8Array(10.1)
if the parameter of the Uint8Array constructor is not integer, chrome browser ignor the 0.1, but the edge browser throw an exception "Invalid offset/length when creating typed array".

cwillisf:
Thanks again, @a49594a! I rearranged your description a little bit in case that helps explain what the different parts of the template are for. I also gave the PR a descriptive name instead of just "Update RenderWebGL.js"

PR #136, opened 2017-04-19
Audit dependency versions

cwillisf:
### Proposed Changes

Remove obsolete dependencies (`json-loader` and `lodash.defaultsdeep`), and minor-pin `json` instead of pinning it to an exact version.

Further versioning changes were proposed and backed out; see the conversation on this PR for details.

This closes #112 

### Reason for Changes

Since `json` is only used as part of the versioning & publishing process, and doesn't directly affect the build output, there's no need to be strict about its version.


package.json
```diff
@@ -22,28 +22,26 @@
     "watch": "webpack --progress --colors --watch --watch-poll"
   },
   "devDependencies": {
-    "babel-core": "^6.23.1",
-    "babel-eslint": "^7.1.1",
-    "babel-loader": "^6.3.2",
-    "babel-polyfill": "^6.22.0",
-    "babel-preset-es2015": "^6.22.0",
-    "base64-loader": "^1.0.0",
```
rschamp: Previously we set these to `^` because babel is particularly spammy re: Greenkeeper, and because we trust babel to stick to the true meaning of semver, so won't release breaking changes here.

package.json
```diff
@@ -22,28 +22,26 @@
...
-    "json-loader": "^0.5.4",
-    "lodash.defaultsdeep": "4.6.0",
-    "raw-loader": "^0.5.1",
+    "json": "^9.0.4",
+    "raw-loader": "0.5.1",
     "scratch-render-fonts": "git+https://github.com/LLK/scratch-render-fonts.git",
     "tap": "^10.3.0",
     "travis-after-all": "^1.4.4",
     "twgl.js": "3.2.2",
-    "webpack": "^2.2.1",
```
rschamp: Similarly webpack will either work or not, and releases quite frequently. Although I don't trust them as much to not release breaking changes. Maybe the webpack-related packages should have `~`?  In our other repos, we pinned babel and webpack-related packages with `^` because the Greenkeeper spam was too much to keep up with (personally).

cwillisf:
Makes sense. One of my motivations for this was to bring out this type of comment. I remembered we talked about pinning things which affect the build output but I don't remember discussing babel or webpack explicitly... it could be that I just forgot, though :P

> In our other repos, we pinned babel and webpack-related packages with ^

This is mostly true for `babel-*`, except in `scratch-storage` which is my "fault" and `scratch-www` which might be treated as a different beast. However, webpack is currently pinned to an exact version everywhere but `scratch-render` :)

cwillisf:
@rschamp 👍
What I was about to say:
> Loaders are an interesting case -- I had been thinking of them more like runtime code but really they're more like webpack, at least in terms of their impact on the build output. However, while some loaders are semi-officially associated with webpack (`raw-loader`), other loaders are written and maintained by individuals (`base64-loader`). I think I'd like to leave them like this for now... we can always change it later if there's specific motivation.

Then I thought about it a little more, and decided that:
1. in practical terms, this probably isn't going to matter, but
2. the "API" on a loader is so locked down that I don't think it's likely to become incompatible even with an "incorrect" semver bump. It seems more likely that we'd miss a bug-fix due to a strictly pinned version number & falling behind on GK PRs.

rschamp:
That all makes sense to me, thanks for thinking it through @cwillisf.

Issue #131, opened 2017-04-10
Rotation center is incorrect while sprite is scaled by "changeSize" block or "setSize" block


a49594a:
### Expected Behavior

Rotation center should stay in the same relative location when sprite is scaled.

### Actual Behavior

The rotation center is not in the middle of the costume, the rotation center changes while sprite is scaled by "changeSize" block or "setSize" block.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

thisandagain:
@cwillisf @rschamp Thoughts on this one?

thisandagain:
Here is a repro case:
https://scratch.mit.edu/projects/156451053/
https://llk.github.io/scratch-gui/#156451053

cwillisf:
Most likely I just got the order of operations wrong in the matrix math in `_calculateTransform`:
https://github.com/LLK/scratch-render/blob/develop/src/Drawable.js#L194

I'm marking this as "help wanted" -- the change probably won't require much knowledge of our codebase, though it does require an understanding of matrix math as applied to transforming geometry.

cwillisf:
Fixed by #138

PR #138, opened 2017-04-22
fix the bug #131 Rotation center is incorrect while sprite is scaled by "changeSize" block or "setSize" block

a49594a:
### Resolves

https://github.com/LLK/scratch-render/issues/131

### Proposed Changes

the "rotationAdjusted" should be scaled while the drawable target is scaled.

rotationAdjusted = twgl.v3.multiply(rotationAdjusted, this.scale);
rotationAdjusted = twgl.v3.divScalar(rotationAdjusted, 100);


thisandagain:
Sorry for the slow reply @a49594a. We'll take a look at this soon.

a49594a:
：)

PR #153, opened 2017-08-07
Fix scratch-render playground

joshlory:
It looks like loading skin by URL is deprecated (see: https://github.com/LLK/scratch-render/commit/64c59914861a47162aee48a8e9f60854118617e8).

### Before

screen shot 2017-08-07 at 3 53 23 pm

### After

screen shot 2017-08-07 at 3 53 07 pm

### DIFF
--- a/src/playground/index.html
+++ b/src/playground/index.html
@@ -44,10 +44,16 @@
         direction: 90
     });
     var drawableID2 = renderer.createDrawable();
-    renderer.updateDrawableProperties(drawableID2, {
-        skin: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
-            '09dc888b0b7df19f70d81588ae73420e.svg/get/'
-    });
+    var image = new Image();
+    image.onload = function () {
+        var skinId = renderer.createBitmapSkin(image);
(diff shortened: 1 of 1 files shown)


src/playground/index.html
```diff
@@ -44,10 +44,16 @@
...
     });
     var drawableID2 = renderer.createDrawable();
-    renderer.updateDrawableProperties(drawableID2, {
-        skin: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
-            '09dc888b0b7df19f70d81588ae73420e.svg/get/'
-    });
+    var image = new Image();
+    image.crossOrigin = 'anonymous';
+    image.src = 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
+        '09dc888b0b7df19f70d81588ae73420e.svg/get/';
```
cwillisf: Just to be safe, please attach the `onload` function before setting `image.src`. It may work as you have it, depending on the browser and cache state, but setting `onload` first is more reliable.
See also: https://stackoverflow.com/questions/14648598/is-it-necessary-to-set-onload-function-before-setting-src-for-an-image-object

joshlory:
Updated!

PR #154, opened 2017-08-09
Fix picking when input size is different from native size

paulkaplan:
### Resolves

Part of being able to resize the stage.

### Proposed Changes

Correctly scale the inputs to the `extractDrawable` function so that the input coords do not have to be the same as the native coords. This happens when you are interacting with a canvas that is at a different size from the "native" size.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -623,15 +623,19 @@ class RenderWebGL extends EventEmitter {
         const drawable = this._allDrawables[drawableID];
         if (!drawable) return null;
 
+        // Convert client coordinates into absolute scratch units
+        const scratchX = this._nativeSize[0] * ((x / this._gl.canvas.clientWidth) - 0.5);
+        const scratchY = this._nativeSize[1] * ((y / this._gl.canvas.clientHeight) - 0.5);
+
         const gl = this._gl;
         twgl.bindFramebufferInfo(gl, this._queryBufferInfo);
 
         const bounds = drawable.getFastBounds();
         bounds.snapToInt();
 
-        // Translate input x and y to coordinates relative to the drawable
-        const pickX = x - ((this._nativeSize[0] / 2) + bounds.left);
-        const pickY = y - ((this._nativeSize[1] / 2) - bounds.top);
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -623,15 +623,19 @@ class RenderWebGL extends EventEmitter {
         const drawable = this._allDrawables[drawableID];
         if (!drawable) return null;
 
+        // Convert client coordinates into absolute scratch units
+        const scratchX = this._nativeSize[0] * ((x / this._gl.canvas.clientWidth) - 0.5);
+        const scratchY = this._nativeSize[1] * ((y / this._gl.canvas.clientHeight) - 0.5);
```
rschamp: What is the `-  0.5`?

paulkaplan: A yeah there is a lot packed in there. `x` starts as the position of the click from the top left in the range `[0, clientWidth]`. Dividing by client width gets you to `[0, 1]`. The `-0.5` then centers the input from `[-0.5, 0.5]`. Then multiplying by the native size brings it to `[-0.5 * nativeSize, 0.5 * nativeSize]` (i.e. scratch coordinates!

src/RenderWebGL.js
```diff
@@ -623,15 +623,19 @@ class RenderWebGL extends EventEmitter {
...
 
         const bounds = drawable.getFastBounds();
         bounds.snapToInt();
 
-        // Translate input x and y to coordinates relative to the drawable
-        const pickX = x - ((this._nativeSize[0] / 2) + bounds.left);
-        const pickY = y - ((this._nativeSize[1] / 2) - bounds.top);
+        // Translate to scratch units relative to the drawable
+        const pickX = scratchX - bounds.left;
+        const pickY = scratchY + bounds.top;
```
rschamp: Here, it looks like the coordinate system for `pickX` and `pickY` is changing? Does this require changes elsewhere?

paulkaplan: Nope, this coordinate system change is actually part of the bug fix.

PR #155, opened 2017-08-10
Add initial test coverage for Drawable.prototype.getAABB

joshlory:
Add a bit of test coverage for `_calculateTransform`.

I have a few additional tests for rotate + scale, but those aren't passing yet.  I'll create a separate PR with those changes!

### Test Coverage

```> scratch-render@0.1.0 tap /Users/joshlory/Work/2k14/scratch-render
> tap ./test/unit/*.js

./test/unit/DrawableTests.js ........................ 10/10
total ............................................... 10/10

  10 passing (466.848ms)

  ok
```

### DIFF
--- a/src/ShaderManager.js
+++ b/src/ShaderManager.js
@@ -1,8 +1,5 @@
 const twgl = require('twgl.js');
 
-const vertexShaderText = require('raw-loader!./shaders/sprite.vert');
-const fragmentShaderText = require('raw-loader!./shaders/sprite.frag');
-
 
 class ShaderManager {
     /**
@@ -65,8 +62,11 @@ class ShaderManager {
         }
 
         const definesText = `${defines.join('\n')}\n`;
-        const vsFullText = definesText + vertexShaderText;
-        const fsFullText = definesText + fragmentShaderText;
+
+        /* eslint-disable global-require */
+        const vsFullText = definesText + require('raw-loader!./shaders/sprite.vert');
+        const fsFullText = definesText + require('raw-loader!./shaders/sprite.frag');
+        /* eslint-enable global-require */
 
         return twgl.createProgramInfo(this._gl, [vsFullText, fsFullText]);
     }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- "4.2"
+- 6
 - "node"
 sudo: false
 cache:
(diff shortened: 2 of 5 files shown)


thisandagain:
Hi @joshlory thanks for adding these tests! Looks like it is still failing CI. Let us know if you have any questions.

joshlory:
The error I'm seeing is:

```
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

Any objections to me changing `.travis.yml` to specify Node 6 instead of Node 4?  I'm looking at these commits:

- https://github.com/LLK/scratch-vm/commit/7dc9fad41e09ea74da6c309ddf880ae716f231b7 Update to Node 6 LTS
- https://github.com/LLK/scratch-gui/commit/24b72f76d3ab952c58e739b8dde37241f9b83130 Update base Node version to 6

thisandagain:
@joshlory Nope. Good catch. Go for it.

joshlory:
Should be fixed now, sorry!

src/ShaderManager.js
```diff
@@ -65,8 +62,11 @@ class ShaderManager {
         }
 
         const definesText = `${defines.join('\n')}\n`;
-        const vsFullText = definesText + vertexShaderText;
-        const fsFullText = definesText + fragmentShaderText;
+
+        /* eslint-disable global-require */
+        const vsFullText = definesText + require('raw-loader!./shaders/sprite.vert');
+        const fsFullText = definesText + require('raw-loader!./shaders/sprite.frag');
+        /* eslint-enable global-require */
```
joshlory: I'm following the pattern from [scratch-storage@8dddd1f](https://github.com/LLK/scratch-storage/blob/8dddd1f8e2554e8f1726c5f768dd7dbc1b06b8b5/src/BuiltinHelper.js#L24).  The issue is that `tap` runs without a compile step, so it doesn't know what to do with `require('raw-loader!...')`.

joshlory:
It might be more correct to rename `rotationCenter` to `costumeCenter`.  Looking at the [Scratch Wiki](https://wiki.scratch.mit.edu/wiki/Costume_Center) (and the behavior in Scratch 2.0) it seems to be the point to base translate, scale & rotate operations.

That's the behavior I've added tests for — let me know if that's not correct!

joshlory:
Ping @cwillisf 🙂.

Issue #158, opened 2017-08-23
Need API for color picking


paulkaplan:
We need to add a public API for getting the color at an XY location. The implementation will be similar to the picking process for sprites. 

The color picking requirements are not stable yet, but it would be nice if the API could allow you to get the exact color at a point AND the pixel image data for a region around the color, so we can implement a zoomed color loupe.

cwillisf:
I recommend looking into [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData) as a possible container for returned pixel data. It stores width, height, and pixels, and it works with the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#Supported_types) so it can cross worker boundaries.

PR #159, opened 2017-08-25
Add color extraction method to renderer

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-render/issues/158

### Proposed Changes

Adds an API similar to `extractDrawable` for picking a color and getting back pixel data from around it. 

### Reason for Changes

Needed for color picking in scratch 3. 

Notes for @cwillisf, we may want to just pair for a little on these:
- Is x, y, radius an ok way to call this method? radius is a bit unusual because it returns a box...
- I didn't use ImageData just because I was sticking closely with the `extractDrawable` code, we should probably update both at the same time.
- The coordinate system stuff is a little strange, probably could be refactored to be more clear. It also doesn't seem to work right with retina screens, but for the same reason as extractDrawable probably doesn't. Maybe we could pair on fixing both of those before merging this?

I also added it as a mousemove effect on the playground, click still does extract drawable. 

simple-eyedropper

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -5,6 +5,7 @@ const twgl = require('twgl.js');
 
 const BitmapSkin = require('./BitmapSkin');
 const Drawable = require('./Drawable');
+const Rectangle = require('./Rectangle');
 const PenSkin = require('./PenSkin');
 const RenderConstants = require('./RenderConstants');
 const ShaderManager = require('./ShaderManager');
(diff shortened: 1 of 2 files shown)


src/RenderWebGL.js
```diff
@@ -685,6 +686,79 @@ class RenderWebGL extends EventEmitter {
...
+        const projection = twgl.m4.ortho(bounds.left, bounds.right, bounds.top, bounds.bottom, -1, 1);
+
+        gl.clearColor(0, 0, 0, 0);
+        gl.clear(gl.COLOR_BUFFER_BIT);
+        try {
+            gl.disable(gl.BLEND);
+            this._drawThese(this._drawList, ShaderManager.DRAW_MODE.default, projection);
+        } finally {
+            gl.enable(gl.BLEND);
+        }
```
cwillisf: I think you probably don't want to disable & re-enable `gl.BLEND` here. That makes sense for `extractDrawable` since it's only operating on one drawable, but here you'll want the blended results of all drawables. Try this with a sprite using the "ghost" effect (even better: use a non-white backdrop) to see what I mean.

Anyway, I think this should probably just be:
```javascript
gl.clearColor(0, 0, 0, 0);
gl.clear(gl.COLOR_BUFFER_BIT);
this._drawThese(this._drawList, ShaderManager.DRAW_MODE.default, projection);
```

src/RenderWebGL.js
```diff
@@ -685,6 +686,79 @@ class RenderWebGL extends EventEmitter {
...
+
+    /**
+     * Return drawable pixel data and color at a given position
+     * @param {int} x The client x coordinate of the picking location.
+     * @param {int} y The client y coordinate of the picking location.
+     * @param {int} radius The client radius to extract pixels with.
+     * @return {?ColorExtraction} Data about the picked color
+     */
+    extractColor (x, y, radius) {
+        // @todo this doesn't seem to work for retina screens
```
cwillisf: Yep - `_nativeSize` is always in terms of pixels (technically texels) whereas `canvas.client{Width,Height}` are in terms of HTML client "pixels" -- which aren't real pixels in HiDPI mode.

It might be a good idea to have this function take "Scratch coordinates" as input and do the conversion from client coordinates at the GUI layer, like this: https://github.com/LLK/scratch-gui/pull/606/files#diff-136323462d986e04bdd73357a9d350c3R71

Don't forget to scale the radius, too... if you're worried about non-square scaling, I think I'd switch this to take a width & height rather than two "radii" but I'm OK with one radius if you're not worried about non-square scaling.

paulkaplan:
@cwillisf paired and updated in 0fbbd68

src/RenderWebGL.js
paulkaplan: I implemented exactly what you have there, and I'm not getting any pixel output anymore... Any idea why?

paulkaplan:
Fixed the issues with CWF. Merging!

PR #160, opened 2017-08-30
Add updateSVGSkin

fsih:
This will allow us to update an existing SVG in the renderer, which is more efficient than making and hooking up a new one. This will be useful for reflecting edits from the paint editor live.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -228,22 +228,18 @@ class RenderWebGL extends EventEmitter {
      * @param {!string} svgData - new SVG to use.
      * @param {?Array<number>} rotationCenter Optional: rotation center of the skin. If not supplied, the center of the
      * skin will be used
-     * @returns {!int} the ID for the skin. This will be different from skinId if the old skin was not SVG.
      */
     updateSVGSkin (skinId, svgData, rotationCenter) {
         if (this._allSkins[skinId] instanceof SVGSkin) {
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -223,6 +223,30 @@ class RenderWebGL extends EventEmitter {
...
+        }
+
+        const newSkinId = this.createSVGSkin(svgData, rotationCenter);
+        for (const drawable of this._allDrawables) {
+            if (drawable && drawable.skin.id === skinId) {
+                drawable.skin = this._allSkins[newSkinId];
+            }
+        }
+        this.destroySkin(skinId);
+        return newSkinId;
```
cwillisf: Instead of this `for` loop, what do you think of something like this?
```js
const newSkin = new SVGSkin(skinId, this);
newSkin.setSVG(svgData, rotationCenter);
const oldSkin = this._allSkins[skinId];
this._allSkins[skinId] = newSkin;
delete oldSkin;
```

fsih: That seems better, then I don't have to worry about propagating the new skin ID in the VM.

Did you mean oldSkin.dispose() instead of delete oldSkin?

cwillisf: Oh, wow... yep, I meant `oldSkin.dispose()`. I guess I had a C++ flashback there...

PR #162, opened 2017-08-31
Match Scratch 2.0 threshold for `isTouchingColor`

joshlory:
### Resolves

https://github.com/LLK/scratch-gui/issues/426: 2.0 project's sprite touching color/ another sprite doesn't trigger; ball falls through floor

### Proposed Changes

Update `isTouchingColor` to use [the same tolerance as Scratch 2.0](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/scratch/ScratchStage.as#L643-L655):
```actionscript
var mask:uint = 0x00F8F8F0;
```

### Reason for Changes

Fixes https://llk.github.io/scratch-gui/#161606350.  Note that the `backdrop9` level still isn't passable in Scratch 3.0.

color-tolerance

### Test Coverage

None 🙁.  I've verified the change manually for the broken project.  In order to run this with `tap` we'd need to mock a lot of twgl.js inside Node.  An alternative might be to run tests with Karma + Chrome Headless.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -29,10 +29,15 @@ const MAX_TOUCH_SIZE = [3, 3];
  * target is touching a color whose components are each within this tolerance of
  * the corresponding component of the query color.
  * between 0 (exact matches only) and 255 (match anything).
- * @type {int}
+ * @type {object.<string,int>}
  * @memberof RenderWebGL
  */
-const TOLERANCE_TOUCHING_COLOR = 2;
+const TOLERANCE_TOUCHING_COLOR = {
+    R: 7,
+    G: 7,
+    B: 15,
+    Mask: 2
+};
 
 /**
  * Sprite Fencing - The number of pixels a sprite is required to leave remaining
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -29,10 +29,15 @@ const MAX_TOUCH_SIZE = [3, 3];
...
  * between 0 (exact matches only) and 255 (match anything).
- * @type {int}
+ * @type {object.<string,int>}
  * @memberof RenderWebGL
  */
-const TOLERANCE_TOUCHING_COLOR = 2;
+const TOLERANCE_TOUCHING_COLOR = {
+    R: 7,
+    G: 7,
+    B: 15,
```
joshlory: Why does Scratch 2.0 specify a different tolerance mask for the blue component (`0xf0`) vs. red & green (`0xf8`)?

Unfortunately the project in the bug report depends on the extra blue tolerance.  The lines are drawn with `#000000` but the code checks for `#00000d`.

cwillisf: To be honest I'm not certain why Scratch 2.0 works this way. I do know that the typical human eye is better at distinguishing minor changes in shades of green than in red or blue, so it might be related to that.

joshlory:
That makes sense!  Feel free to close this PR, I'm not invested either way 😄.

cwillisf:
Nah, I think this change is good. The tolerance we had was too strict - even more strict than the theoretical average case in Scratch 2.0 - so it makes sense to make it a little less strict. Thanks for the change!

PR #167, opened 2017-09-06
chore(package): update twgl.js to version 3.5.0

cwillisf:
Closes #141


cwillisf:
Tested locally, no problems encountered

PR #176, opened 2017-10-04
Add utils for text wrapping, originally by CWF

paulkaplan:
### Proposed Changes

Add utilities for text wrapping to the renderer, in preparation for say/think bubbles. This is a mirror of the original PR @cwillisf made to GUI, with just a minor adjustment for font sizing that I found while implementing say/think. 

@cwillisf this is meant as a starting point for us to discuss the `TODO` mentioned in the code.


src/util/svg-text-wrapper.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+class SVGMeasurementProvider {
+    constructor () {
+        this._svgRoot = null;
+    }
+
+    /**
+     * Detach the hidden SVG element from the DOM and forget all references to it and its children.
+     */
+    dispose () {
```
paulkaplan: @cwillisf I noticed this method is never called by the text wrapper and also isn't called in the `endMeasurementSession` method. Seems like maybe a relic of a previous API? Where should it be called?

(Found this out when I was poking around the DOM and noticed that there were a bunch of suspicious SVG's attached to the end of the DOM!)

src/util/svg-text-wrapper.js
```diff
@@ -0,0 +1,99 @@
...
+     */
+    _init () {
+        const svgNamespace = 'http://www.w3.org/2000/svg';
+
+        const svgRoot = document.createElementNS(svgNamespace, 'svg');
+        const svgGroup = document.createElementNS(svgNamespace, 'g');
+        const svgText = document.createElementNS(svgNamespace, 'text');
+
+        // Normalize text attributes to match what the svg-renderer does.
+        // @TODO This code should be shared with the svg-renderer.
```
paulkaplan: @cwillisf so for this PR, do you think I should:
(a) Leave this the way it is
(b) Extract the text formatting part of the svg renderer into a utility and use it both here and in the svg renderer.
(c) Something else?

paulkaplan:
@cwillisf I updated this PR with a few tweaks:
- Fixed the `#dispose` method so that it is actually called. Let me know if that looks right.
- Added caching for text measurements so that we wouldn't have to keep going to the DOM for repeat strings measurement.

src/util/svg-text-wrapper.js
cwillisf: What I had in mind is that this would be called only when the `SVGMeasurementProvider` is never going to be used again -- not after every measurement session. I probably should have hooked it up to a `dispose()` method on `SVGTextWrapper`, to be called by whoever owns the text wrapper.

Calling `dispose()` at the end of each measurement session is OK for now I think, but I imagined adding & removing an SVG node on the DOM to be a relatively expensive operation so I wanted to minimize the number of times it happens. [Premature optimization](http://wiki.c2.com/?PrematureOptimization), maybe :)

PR #177, opened 2017-10-04
Say/think bubble SVG templates

paulkaplan:
### Proposed Changes

This PR builds on https://github.com/LLK/scratch-render/pull/176 (the first commit here) and two parts: 

1. A utility for generating say/think bubble svg strings. 
2. A public API for creating skins using the say/think bubble svg templates by passing in the text to be displayed and some other required info. 

This is meant as conversation-starter code, it is still very early. The things I want to talk about w/ @cwillisf are:

- [x] This code hijacks the `svg-renderer` in order to measure the width/height of the text block before creating the enveloping path. The `svg-renderer` isn't meant for this, because it also draws it to a canvas which we do not need. Is there a better way to share the measurement code, possibly drawing out and expanding the `SVGMeasurementProvider` used in the text wrapping code?
- [x] This code exposes new methods for creating/update "text skins", but there is no TextSkin subclass, it is really just an SVG skin. Is that an ok way of doing it?


src/util/svg-text-bubble.js
```diff
@@ -0,0 +1,176 @@
...
+              ${fragment}
+          </svg>
+        `;
+    }
+
+    _textFragment () {
+        return `<text fill="#575E75">${this.lines.join('\n')}</text>`;
+    }
+
+    toString (type, text, pointsLeft) {
```
cwillisf: I'm a bit uncomfortable calling this `toString`... I think I see what you were going for but this isn't really a conversion operation, which is what the `to` says to me. Since it's building something pretty much from scratch, I think I'd use a verb like `make` or `build`.

src/util/svg-text-bubble.js
```diff
@@ -0,0 +1,176 @@
...
+            </g>`;
+    }
+
+
+    _getTextSize () {
+        const svgString = this._wrapSvgFragment(this._textFragment());
+        if (!this._textSizeCache[svgString]) {
+            this._textSizeCache[svgString] = this.svgRenderer.measure(svgString);
+        }
+        return this._textSizeCache[svgString];
```
cwillisf: Considering that the measurement provider also caches its results, this feels almost redundant to me. I recognize that it's a bit different, though. Maybe as a future enhancement we can find a way to only cache in one place.

src/util/svg-text-bubble.js
paulkaplan: Good idea. Changed in 04e9d6c

src/util/svg-text-bubble.js
paulkaplan: Yeah, unfortunately they are slightly different. I went back and forth trying to figure out a way to combine them, which would also help because we are measuring nearly the same thing twice in the non-cached case. But yeah i think that would be a good future enhancement.

Issue #175, opened 2017-10-03
Clipping of SVG strokes 


paulkaplan:
### Expected Behavior

_Please describe what should happen_

![screen shot 2017-10-03 at 8 57 59 am](https://user-images.githubusercontent.com/654102/31126243-17f61b2c-a819-11e7-88ab-f858e2da1361.png)

### Actual Behavior

_Describe what actually happens_

![screen shot 2017-10-03 at 8 57 47 am](https://user-images.githubusercontent.com/654102/31126251-1c4bcf32-a819-11e7-8054-4a89b509547e.png)

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

Those screenshots are from importing a simple SVG with a somewhat thicker stroke width set. This is the SVG I used:
```
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg" version="1.1">
    <circle r="50" cx="50" cy="50" stroke-width="5" stroke="black" fill="none" />
</svg>
```
I have seen this in a number of real-world projects as well, for example https://scratch.mit.edu/projects/22508123/. This also came up when trying to design think/say bubbles in SVG. They get clipped in the same way.

We use `svg.getBBox()` to calculate the width/height of the svg, which only uses the geometry of the elements and not visual styles like stroke-width ([source](https://stackoverflow.com/a/34340922)). I'm not sure how we can solve this...

fsih:
@paulkaplan is your original example still fixed on develop Paul? Not sure if this regressed or is a different version of this bug, but on https://llk.github.io/scratch-gui/develop/#180574519 the clipping issue still exists.

PR #179, opened 2017-10-05
Enlarge bounding box by the largest stroke width.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-render/issues/175

### Proposed Changes

Enlarge the bounding box that is read from the DOM method `getBBox()` which doesn't take into account the strokes by the largest found stroke width. As mentioned in the code comment, this may have false positives, as in it may enlarge an SVG viewbox that didn't need it (if, for example, the largest shape was a fill, not a stroke). But it will always make the viewbox large enough to include the strokes, which seems like an ok trade-off. 

### Reason for Changes

Clipping of SVG's, which is impacting project rendering and also messes with the upcoming speech bubble rendering.

### Test Coverage

None.

/cc @tmickel since it looks like a lot of this code was yours.

### DIFF
--- a/src/svg-quirks-mode/svg-renderer.js
+++ b/src/svg-quirks-mode/svg-renderer.js
@@ -163,6 +163,34 @@ class SvgRenderer {
     }
 
     /**
+     * Find the largest stroke width in the svg. If a shape has no
+     * `stroke` property, it has a stroke-width of 0. If it has a `stroke`,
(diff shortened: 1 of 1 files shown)


fsih:
Should it be sufficient to enlarge by 1/2 the largest stroke width?

paulkaplan:
@fsih good catch! done in 
ab4046b

PR #188, opened 2017-11-07
Compare new setRotationCenter values after truncating to 32bit

mzgoddard:
### Resolves

Related to https://github.com/LLK/scratch-vm/issues/738

### Proposed Changes

Truncate new rotation center values to 32 bit before comparing against current values.

### Reason for Changes

`Skin._rotationCenter` is a twgl vector or said another way a Float32Array. JavaScript casts values stored in this array as 32 bit floats instead of JavaScript's standard 64 bit floats. Comparing the
equality, like `setRotationCenter` does, of a 64 bit float against a value stored in a 32 bit container will virtually always be false. Truncating the new value into a 32 bit (losing precision) will give a
more useful comparison and remove a lot of current false positives.

This greatly improves execution time for projects creating clones and using the `if on edge, bounce` block like: https://scratch.mit.edu/projects/130041250/

### Test Coverage

Fixing existing functionality.

### DIFF
--- a/src/Skin.js
+++ b/src/Skin.js
@@ -4,6 +4,19 @@ const twgl = require('twgl.js');
 
 const RenderConstants = require('./RenderConstants');
 
+/**
+ * Truncate a number into what could be stored in a 32 bit floating point value.
+ * @param {number} num Number to truncate.
+ * @return {number} Truncated value.
+ */
+const toFloat32 = (function () {
+    const memory = new Float32Array(1);
+    return function (num) {
+        memory[0] = num;
+        return memory[0];
+    };
+}());
+
 class Skin extends EventEmitter {
     /**
      * Create a Skin, which stores and/or generates textures for use in rendering.
@@ -79,7 +92,11 @@ class Skin extends EventEmitter {
      */
     setRotationCenter (x, y) {
         const emptySkin = this.size[0] === 0 && this.size[1] === 0;
-        const changed = x !== this._rotationCenter[0] || y !== this._rotationCenter[1];
+        // Compare a 32 bit x and y value against the stored 32 bit center
+        // values.
+        const changed = (
+            toFloat32(x) !== this._rotationCenter[0] ||
+            toFloat32(y) !== this._rotationCenter[1]);
         if (!emptySkin && changed) {
             this._rotationCenter[0] = x;
             this._rotationCenter[1] = y;


mzgoddard:
How this affects an example project in more detail:

https://scratch.mit.edu/projects/130041250/ creates 50 clones when it starts. Each clone creates a drawable, updates all drawable properties, including the skin's rotation center. The new rotation center values do not fit into 32 bit floats and when compared against the last set values are different. The "new" values are stored into `_rotationCenter` after JavaScript automatically truncates them into 32 bit values. Changing the rotationCenter emits a Skin.Events.WasAltered. All the current Drawables listenging to that skin's WasAltered invalidate their convex hull. Later the all the current Target's run `if on edge, bounce` which uses the Drawable's getBounds called through the renderer. The drawable has to rebuild its convex hull. On my machine building the hull takes 20 to 30ms.

Until all the clones are created, all existing clones are rebuild there hull every runtime step. With this fix only the most recent clone needs to build its hull. This means the step creating the last clone goes from taking **1050 milliseconds** on my machine to **138 milliseconds**. All the cloning frames together goes from **29.2 seconds** to **4.65 seconds**.

src/Skin.js
```diff
@@ -79,7 +92,11 @@ class Skin extends EventEmitter {
      */
     setRotationCenter (x, y) {
         const emptySkin = this.size[0] === 0 && this.size[1] === 0;
-        const changed = x !== this._rotationCenter[0] || y !== this._rotationCenter[1];
+        // Compare a 32 bit x and y value against the stored 32 bit center
+        // values.
+        const changed = (
+            toFloat32(x) !== this._rotationCenter[0] ||
+            toFloat32(y) !== this._rotationCenter[1]);
```
cwillisf: What do you think about putting `x = toFloat32(x);` and `y = toFloat32(y);` near the top instead? I suppose the truncation itself isn't a huge expense, but with this formulation I think we'll usually end up doing more conversions than necessary (counting implicit conversions in the assignments below).

| Conversion count | `toFloat32` in `changed` | `x=toFloat32(x)`  |
-|-|-
| x and y unchanged | 2 | 2 |
| x changed, y same | 3 | 2 |
| x same, y changed | 4 | 2 |
| x and y changed | 3 | 2 |

src/Skin.js
```diff
@@ -4,6 +4,19 @@ const twgl = require('twgl.js');
 
 const RenderConstants = require('./RenderConstants');
 
+/**
+ * Truncate a number into what could be stored in a 32 bit floating point value.
+ * @param {number} num Number to truncate.
+ * @return {number} Truncated value.
+ */
+const toFloat32 = (function () {
+    const memory = new Float32Array(1);
```
cwillisf: 😎

Issue #192, opened 2017-11-13
Improve `_getConvexHullPointsForDrawable` performance


mzgoddard:
### Expected Behavior

Take less than a millisecond, or better, to create a convex set of points containing the drawable's shape.

### Actual Behavior

`_getConvexHullPointsForDrawable` currently has performance like [`pick`](). It renders the drawable to the webgl buffer and reads back with `readPixels` which busy waits for the GPU to stop what its doing and read out the requested state to the CPU. This takes milliseconds to 10s of milliseconds depending on the running hardware.

### Steps to Reproduce

Open a project that uses `if on edge, bounce` like https://scratch.mit.edu/projects/130041250/. When a drawable is created or any change to the drawable invalidates the convex hull, and its owning target executes `if on edge, bounce` it needs to build a new convex hull. The linked example project encounters this issue while it creates 50 clones at the beginning.

Before https://github.com/LLK/scratch-render/pull/188 is merged, its especially clear the `_getConvexHullPointsForDrawable` performance is like. As it creates the clones, the fixed bug in #188 causes every clone to rebuild its hull every frame a clone is created. Leading up to 50 clones, the last frame can lapse into a full second for some hardware.

### Operating System and Browser

All

mzgoddard:
Related to https://github.com/LLK/scratch-vm/issues/738

mzgoddard:
Building on https://github.com/LLK/scratch-render/issues/191#issuecomment-342862720, `pick` and `_getConvexHullPointsForDrawable` could work with a Shape class that contains the alpha channel of a skin. With an alpha channel Shape, the Drawable's model transform, and the effect texture coordinate math in the shader (rewritten in js with twgl), `_getConvexHullPointsForDrawable` can build the points on the CPU with JS and not have to wait for the GPU to stop itself and read back to us.

rschamp:
Fixed in #196

PR #196, opened 2017-11-15
Emulate silhouette rendering for pick and _getConvexHullPointsForDrawable

### LINKED ISSUE #191: Improve `pick` performance
### Expected Behavior

Take less than a millisecond, or better, to pick a drawable from a given x, y position.

### Actual Behavior

On a Samsung Chromebook Plus a single pick can take 10s of milliseconds to execute. (Hard to tell actual performance cause the calls take so long it can bleed into the recorded time for the next pick operation.) The underlying `readPixels` method is also known to have a poorly defined execution time due to its behaviour.

### Steps to Reproduce

Projects like https://scratch.mit.edu/projects/130041250/ use pick to test if the Drawable is touching the mouse-pointer. Once that project has created all its clones, each clone tests if its touching the mouse-pointer, leading to frame times from **250 ms** (milliseconds) to **350 ms**.

All

mzgoddard:
### Resolves

#191 #192

Related to https://github.com/LLK/scratch-vm/issues/738

### Proposed Changes

Emulate silhouette rendering with Silhouette and EffectTransform types.

RenderWebGL, Drawable, and Skins use these types to take world and texture coordinates and query the opacity of a point on a Skin instance. Silhouette uses a texture coordinate to lookup a stored cell in its data to determine if the point on the Skin is opaque or translucent. EffectTransform transforms a texture coordinate into another texture coordinate based on a Drawable's effect values. Drawable.isTouching turns a world coordinate into a drawable coordinate and then into a texture coordinate, transforms that into another texture coordinate with EffectTransform and tests the final coordinate with `this.skin.isTouching`.

`pick` uses Drawable.isTouching in place of its past silhouette rendering and reading.

`_getConvexHullPoitnsForDrawable` uses Skin.isTouching and EffectTransform.transformPoint in place of its past silhouette rendering and reading.

#### Terms

I don't write these terms as official ones but to explain their use above.

**world coordinate** A twgl.v3 object with x and y values in scratch's common -240 to 240 x constraint and -180 to 180 y constraint.
**drawable coordinate** A twgl.v3 object with a x value between -0.5 and 0.5 and a y value between -0.5 and 0.5.
**texture coordinate** A twgl.v3 object with a x value between 0 and 1 and a y value between 0 and 1. Unbounded texture coordinates for isTouching methods return false.

### Reason for Changes

`pick` and `_getConvexHullPointsForDrawable` render drawable silhouettes with webgl and read back the pixels. Without combining the silhouette/touching renderings of RenderWebGL into one render per step, `gl.readPixels` will be a performance blocker while it busy waits for the WebGL context to complete its work and return the produced pixel info.

https://scratch.mit.edu/projects/130041250/ as an example case with usage of the above methods, starts by cloning a target 50 times. Each uses `if on edge, bounce` which builds a bounding box with a convex hull. Without a change like https://github.com/LLK/scratch-render/pull/188, each cloning causes all existing clones to rebuild their convex hull. Adding up that can cause systems like a Samsung Chromebook Plus to take up to **a second (<1fps)** for many of the last steps as it approaches 50 clones.

Each clone also checks if they are currently touching the mouse. So even after the 50 clones are no longer rebuilding their convex hull, they each use `RenderWebGL.pick` to see if they are touching the mouse pointer. The same Samsung Chromebook Plus takes anywhere from **180ms to 300ms (~2-12fps)** for each step.

Testing with this PR's change has shown the Samsung Chromebook Plus can run the same project at **~30fps** while cloning (without https://github.com/LLK/scratch-render/pull/188) and an easy **60fps** after.

### Test Coverage

Existing tests don't yet cover `pick` and `_getConvexHullPointsForDrawable`.

### Further Thoughts

This work could probably be extended in later work to also emulate the silhouette and color testing in methods like `RenderWebGL.isTouchingColor` with a new type called something like `ColorBuffer` which performs behaviour like `Silhouette` but returns color information instead of booleans. If we need to support another rendering behaviour like Multi-Sampling Anti-Aliasing (MSAA) we could do that adding support for that in ColorBuffer where the concern would probably best sit.

`_getConvexHullPointsForDrawable` still takes something like 1ms to run for the blocks in https://scratch.mit.edu/projects/130041250/ on a Samsung Chromebook Plus. Some further strategies could be looked into and probably added to Silhouette to cache `boundaryPoints` data. Possibly looping counter clockwise over every row and column finding the "left" most point that isn't empty. Another option may be to build less accurate hulls by stepping over a downscaled version of the same Silhouette data.

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -4,7 +4,9 @@ const Rectangle = require('./Rectangle');
 const RenderConstants = require('./RenderConstants');
 const ShaderManager = require('./ShaderManager');
 const Skin = require('./Skin');
+const EffectTransform = require('./EffectTransform');
 
+const __isTouchingPosition = twgl.v3.create();
 
 class Drawable {
     /**
@@ -49,6 +51,8 @@ class Drawable {
         this._scale = twgl.v3.create(100, 100);
         this._direction = 90;
         this._transformDirty = true;
+        this._inverseMatrix = twgl.m4.create();
+        this._inverseTransformDirty = true;
         this._visible = true;
         this._effectBits = 0;
 
@@ -73,6 +77,7 @@ class Drawable {
      */
     setTransformDirty () {
         this._transformDirty = true;
+        this._inverseTransformDirty = true;
     }
 
     /**
@@ -239,6 +244,50 @@ class Drawable {
     setConvexHullPoints (points) {
         this._convexHullPoints = points;
         this._convexHullDirty = false;
+    }
+
+    /**
+     * Check if the world position touches the skin.
+     * @param {twgl.v3} vec World coordinate vector.
+     * @return {boolean} True if the world position touches the skin.
+     */
+    isTouching (vec) {
+        if (!this.skin) {
+            return false;
+        }
+
+        if (this._transformDirty) {
+            this._calculateTransform();
+        }
+
+        // Get the inverse of the model matrix or update it.
+        const inverse = this._inverseMatrix;
+        if (this._inverseTransformDirty) {
+            const model = twgl.m4.copy(this._uniforms.u_modelMatrix, inverse);
+            // The normal matrix uses a z scaling of 0 causing model[10] to be
+            // 0. Getting a 4x4 inverse is impossible without a scaling in x, y,
+            // and z.
+            model[10] = 1;
+            twgl.m4.inverse(model, model);
+            this._inverseTransformDirty = false;
+        }
+
+        // Transfrom from world coordinates to Drawable coordinates.
+        const localPosition = twgl.m4.transformPoint(inverse, vec, __isTouchingPosition);
+
+        // Transform into texture coordinates. 0, 0 is the bottom left. 1, 1 is
+        // the top right.
+        localPosition[0] += 0.5;
+        localPosition[1] += 0.5;
+        // The RenderWebGL quad flips the texture's X axis. So rendered bottom
+        // left is 1, 0 and the top right is 0, 1. Flip the X axis so
+        // localPosition matches that transformation.
+        localPosition[0] = 1 - localPosition[0];
+
+        // Apply texture effect transform.
+        EffectTransform.transformPoint(this, localPosition, localPosition);
+
+        return this.skin.isTouching(localPosition);
     }
 
     /**
--- a/src/EffectTransform.js
+++ b/src/EffectTransform.js
@@ -1,0 +1,102 @@
+/**
+ * @fileoverview
+ * A utility to transform a texture coordinate to another texture coordinate
+ * representing how the shaders apply effects.
+ */
+
+const twgl = require('twgl.js');
+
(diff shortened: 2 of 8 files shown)


src/Drawable.js
```diff
@@ -242,6 +247,47 @@ class Drawable {
...
+            twgl.m4.inverse(model, model);
+            this._inverseTransformDirty = false;
+        }
+
+        // Transfrom from world coordinates to Drawable coordinates.
+        const localPosition = twgl.m4.transformPoint(inverse, vec, __isTouchingPosition);
+
+        // Transform into texture coordinates.
+        localPosition[0] += 0.5;
+        // Invert the x axis.
```
cwillisf: I would add a comment here indicating why the X axis must be inverted

src/EffectTransform.js
```diff
@@ -0,0 +1,102 @@
...
+
+class EffectTransform {
+    /**
+     * Transform a texture coordinate to one that would be select after applying shader effects.
+     * @param {Drawable} drawable The drawable whose effects to emulate.
+     * @param {twgl.v3} vec The texture coordinate to transform.
+     * @param {?twgl.v3} dst A place to store the output coordinate.
+     * @return {twgl.v3} The coordinate after being transform by effects.
+     */
+    static transformPoint (drawable, vec, dst) {
```
cwillisf: I'm not a huge fan of logically (though not quite literally) duplicating all this shader code here, but I guess it's worth it considering how much of a performance improvement we get from doing so. Still, I'd like to revisit this post-release.

src/RenderWebGL.js
```diff
@@ -1152,63 +1141,70 @@ class RenderWebGL extends EventEmitter {
...
+        const CCW = function (A, B, C) {
+            return (((B[0] - A[0]) * (C[1] - A[1])) - ((B[1] - A[1]) * (C[0] - A[0])));
         };
-        for (let y = 0; y <= height; y++) {
+
+        // https://github.com/LLK/scratch-flash/blob/dcbeeb59d44c3be911545dfe54d
+        // 46a32404f8e69/src/scratch/ScratchCostume.as#L369-L413 Following
+        // RasterHull creation, compare and store left and right values that
+        // maintain a convex shape until that data can be passed to `hull` for
+        // further work.
```
cwillisf: I'm not sure anyone on the team understands this algorithm any more; it looks like it was copied into the older Scratch codebase from Numerical Recipes or something. Anyway, I guess this is another "let's revisit this post-release": I think we should try to make this code readable to the average Scratch open-source contributor.

mzgoddard: I commented the code below as a first step to make it more understandable.

src/EffectTransform.js
mzgoddard: Same.

src/RenderWebGL.js
cwillisf: Awesome - thanks!

PR #193, opened 2017-11-14
Touching color white always returns "true"

griffpatch:
Speculative fix for https://github.com/LLK/scratch-vm/issues/710

I've submitted this pull request, but I am very sorry to say I am completely unable to test my own fix as I am unable to get scratch-render to compile under windows.

Please do not therefore pull this into develop until someone can varified that it works and has no side effects? Thanks.

This is the error I get when I try to build:

ERROR in ./node_modules/grapheme-breaker/src/GraphemeBreaker.js
    Module not found: Error: Can't resolve 'fs' in 'C:\Users\andrew.griffin\Documents\GitHub\scratch-render\node_modules\grapheme-breaker\src'
     @ ./node_modules/grapheme-breaker/src/GraphemeBreaker.js 9:7-20

PR #197, opened 2017-11-16
Fix ify-loader test on Windows / Cygwin

cwillisf:
### Resolves

This resolves a Windows-specific build problem mentioned by @griffpatch in #193 

### Proposed Changes

This change alters the regex used to determine whether or not to use `ify-loader` for a particular JS source file. The regex is altered such that either forward or backslashes are accepted as path separators, where previously only forward slashes were recognized.

### Reason for Changes

Without this change, attempting to build on Windows will fail in most configurations.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -22,7 +22,7 @@ const base = {
                 }
             },
             {
-                test: /node_modules\/(linebreak|grapheme-breaker)\/.*\.js$/,
+                test: /node_modules[\\/](linebreak|grapheme-breaker)[\\/].*\.js$/,
                 loader: 'ify-loader'
             }
         ]


paulkaplan:
@fsih if you have a minute, could you try this on your windows machine to confirm you can build scratch render?

griffpatch:
Just to note, it worked for me and I can now debug after this change.

I do get another issue in windows which is:

    ERROR in scratch-vm.min.js from UglifyJs
    Unexpected token: name (urlParts) [scratch-vm.min.js:16903,4]

I'm pretty sure this is unrelated though but was curious if another user
was attempting to build in windows whether they also encounter this one.

paulkaplan:
I actually just saw that also, on a mac, and saw it recently on a different repo that also used webpack-dev-server, i believe it is this issue: https://github.com/LLK/scratch-vm/issues/774. I'll file a bug here also

fsih:
I'm also using Linux on Windows. Actually I don't really know how to set up our repos on Windows directly, though I think cwillisf does it using Cygwin.

Linux on windows is a dev feature of Windows. You can turn it on if you go to "Control Panel > Turn Windows features on or off" and check off "Windows subsystem for Linux", then follow the directions at https://msdn.microsoft.com/en-us/commandline/wsl/install-win10#for-anniversary-update-and-creators-update-install-using-lxrun

cwillisf:
Before WSL came out I generally used Cygwin, but would sometimes check plain ol' `cmd`. It gets complicated since many of the tools can be installed as a Cygwin version or a Windows-native version or both. In my case I used to use two combinations:
- Cygwin `bash` prompt, Cygwin git, Windows-native Node.js
- `cmd` prompt, MinGW git ("git for Windows"), Windows-native Node.js

Now I just use Ubuntu through WSL, installed with `lxrun` as @fsih mentioned. If you have the Fall Creators Update, which started rolling out last month, then you can install a slightly better version of WSL through the Windows Store instead of using `lxrun`.

Issue #198, opened 2017-11-16
In range update for webpack-dev-server does not work


paulkaplan:
I believe this issue from the VM is also impacting the renderer repo: https://github.com/LLK/scratch-vm/issues/774

Reported by @griffpatch https://github.com/LLK/scratch-render/pull/197#issuecomment-344961933

PR #199, opened 2017-11-17
Fix webpack dev server

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-render/issues/198

Identical fix as https://github.com/LLK/scratch-vm/pull/792

### Proposed Changes

Only uglify in production

### Reason for Changes

To fix the local development using webpack-dev-server.

cc @griffpatch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 node_js:
 - 6
 - "node"
+env:
+  - NODE_ENV=production
 sudo: false
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,6 +2,8 @@ language: node_js
 node_js:
 - 6
 - "node"
+env:
+  - NODE_ENV=production
```
rschamp: Since `devDependencies` aren't installed with `NODE_ENV=produciton`, you'll need to adjust the `install` step to include `--production=false` to make sure everything that's needed to run the tests is installed.

PR #204, opened 2017-11-30
Use bilinear filtering except when we shouldn't

### LINKED ISSUE #77: Pixely images rendered
### Expected Behavior

Image should not be pixely. 

### Actual Behavior

screenshot 2017-01-27 at 8 57 49 pm

### Steps to Reproduce

1. In this case, I opened Scratch GUI but I think it is a rendering issue...
2. Observe the already place scratch cat

Chromebook latest version

---
Moved from LLK/scratch-vm#405
/cc @CosmicWebServices

cwillisf:
### Resolves

This resolves #77 and generally makes rendering look a bit better in many situations.

### Proposed Changes

Use bilinear filtering unless the `Drawable` is angled by a multiple of 90 degrees and is either a raster image (like a `BitmapSkin`) or is around its native resoluion. This makes both bitmaps and vectors look better in most cases. This is roughly the same logic that Scratch 2.0 used; we may want to tweak it over time to see if we can make it even better.

### Reason for Changes

Bilinear filtering helps with the "jaggy" appearance that sprites sometimes have, especially bitmap images when rotated to non-right angles (wrong angles?). On the other hand, Scratch has a lot of pixel art that looks fuzzy when bilinear filtering is enabled. This logic offers a decent compromise: use nearest-neighbor filtering when the image is capable of looking sharp on its own, and switch to bilinear filtering when nearest-neighbor is likely to look bad.

### Test Coverage

We have no automated tests for image quality. I set up project 143486954 to help with manual testing of this issue. Press 1, 2, or 3 to toggle the pixelate effect, 45 degree rotation, or 150% scale, respectively, and click the sprite to switch between its costumes.

The main thing to watch out for is a result like this (pixelate and scale on, rotate off). I've tested locally in both high-DPI and normal-DPI screen resolutions and haven't seen this problem with these changes in place.
image

### DIFF
--- a/src/BitmapSkin.js
+++ b/src/BitmapSkin.js
@@ -40,6 +40,13 @@ class BitmapSkin extends Skin {
     }
 
     /**
+     * @returns {boolean} true for a raster-style skin (like a BitmapSkin), false for vector-style (like SVGSkin).
+     */
+    get isRaster () {
+        return true;
+    }
+
+    /**
      * @return {Array<number>} the "native" size, in texels, of this skin.
      */
     get size () {
@@ -71,11 +78,9 @@ class BitmapSkin extends Skin {
             gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, bitmapData);
             this._silhouette.update(bitmapData);
         } else {
+            // TODO: mipmaps?
             const textureOptions = {
                 auto: true,
-                mag: gl.NEAREST,
-                /** @todo mipmaps, linear (except pixelate) */
-                min: gl.NEAREST,
                 wrap: gl.CLAMP_TO_EDGE,
                 src: bitmapData
             };
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -902,7 +902,6 @@ class RenderWebGL extends EventEmitter {
      * @return {Array.<number, number>} The fenced position as an array [x, y]
      */
     getFencedPositionOfDrawable (drawableID, position) {
-
         let x = position[0];
         let y = position[1];
 
@@ -1096,7 +1095,6 @@ class RenderWebGL extends EventEmitter {
      * @private
      */
     _drawThese (drawables, drawMode, projection, opts = {}) {
-
         const near = function (a, b, relativeTolerance = 0.01) {
             const absA = Math.abs(a);
             const absB = Math.abs(b);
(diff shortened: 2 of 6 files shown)


src/BitmapSkin.js
```diff
@@ -71,11 +78,9 @@ class BitmapSkin extends Skin {
             gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, bitmapData);
             this._silhouette.update(bitmapData);
         } else {
+            // TODO: mipmaps?
```
cwillisf: A mipmapped texture occupies 4/3 as much memory as a non-mipmapped texture, but usually looks better and can improve performance on weaker graphics chips.

src/RenderWebGL.js
```diff
@@ -1096,6 +1096,14 @@ class RenderWebGL extends EventEmitter {
...
      */
     _drawThese (drawables, drawMode, projection, opts = {}) {
+
+        const near = function (a, b, relativeTolerance = 0.01) {
+            const absA = Math.abs(a);
+            const absB = Math.abs(b);
+            const error = Math.abs(a - b) / Math.max(absA, absB);
+            return error < relativeTolerance;
+        };
+
```
cwillisf: I wrote this function when I was attempting to compare dimensions instead of scaling factors... it might be overly complicated now but it does make it clear what I'm doing, so... opinions?

src/RenderWebGL.js
```diff
@@ -1096,6 +1096,14 @@ class RenderWebGL extends EventEmitter {
      * @private
      */
     _drawThese (drawables, drawMode, projection, opts = {}) {
+
```
thisandagain: Nit: extra line break

src/BitmapSkin.js
thisandagain: This would be good a file a follow-up issue for rather than tracking in `@todo`s.

thisandagain:
@cwillisf Did you do any performance testing with this? If so, any impact?

cwillisf:
@thisandagain, I didn't do any formal performance testing, but bilinear filtering is generally treated as free (or very close to free) on parallelized GPUs -- that's just about any GPU from ~2000 or later.

On the JS side of things, I would expect `near` to be relatively cheap: 2 calls to `abs`, one to `max`, and a divide, but no GC impact. I also expect that the logic for `useNearest` would be relatively cheap, being only one modulo and a bit of Boolean logic, plus a branch in the next line. On the other hand, I expect the changes I made to the handling of uniforms will actually decrease the GC impact of `_drawThese` by slightly decreasing the number of temporary objects created, so overall I'd expect this PR to be a small win.

All that said, I don't notice any positive or negative performance impact on the devices I have available to me.

thisandagain:
Thanks for the extra analysis @cwillisf. This looks good to land for me.

Issue #205, opened 2017-11-30
_getConvexHullPointsForDrawable fails for some costumes?


paulkaplan:
```
Uncaught TypeError: Cannot read property '0' of undefined
    at CCW (lib.min.js:176824)
    at RenderWebGL._getConvexHullPointsForDrawable (lib.min.js:176861)
    at RenderWebGL.getBounds (lib.min.js:176011)
    at RenderedTarget.getBounds (lib.min.js:7990)
    at Scratch3LooksBlocks._renderBubble (lib.min.js:18642)
    at Scratch3LooksBlocks._updateBubble (lib.min.js:18675)
    at Scratch3LooksBlocks.say (lib.min.js:18725)
    at Scratch3LooksBlocks.sayforsecs (lib.min.js:18732)
    at execute (lib.min.js:22908)
    at Sequencer.stepThread (lib.min.js:25371)
```

To repro:
1.  Go to https://llk.github.io/scratch-gui/develop/
2. Add the sprite "Dinosaur2"
3. Click to activate one of the "say" blocks.
4. ... above error.

Seems to be sprite dependent? 
Works: default cat, "Cat Flying", "Beachball"
Doesn't work: "Duck", "Dinosaur2"

You can also get this happening in the playground, but you need to change to loading an svg skin instead of a bitmap one. Replace the drawable2 section of the playground with this:
```
    var drawableID2 = renderer.createDrawable();
    fetch('https://cdn.assets.scratch.mit.edu/internalapi/asset/a01d31caaab3b14b9bb0e2cd5a86c70c.svg/get/').then(function (response) {
        return response.text();
    }).then(function(svg) {
        var skinId = renderer.createSVGSkin(svg, [0, 0]);
        renderer.updateDrawableProperties(drawableID2, {
            skinId: skinId
        });
    });
```

And then I'm able to call `renderer.getBounds(drawableID2)` and it blows up the same way.

paulkaplan:
Ok investigated further. It seems to be due to this line https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L1208-L1210 which implicitly assumes that the width coming from `drawable.size.width` is an integer. However, that does not seem to be the case. In the case where the scan does not find a touchable pixel on the first y iteration, and the size is not an integer, the incremented "x" value ends up for example at 140 where the width is 139.72, which allows the rest of the code to execute, causing the other problems.

[...]

PR #206, opened 2017-12-01
Fix convex hull calculation bug

paulkaplan:
Fixes a bug with the convex hull calculation where the bounds could not
be calculated for certain costumes with non-integer sizes.

### Resolves

Resolves https://github.com/LLK/scratch-render/issues/205

### Proposed Changes

Because `width` can be non-integer, check if x has gotten to or surpassed the width value while incrementing.

---
@cwillisf @mzgoddard as I mentioned in the linked issue, this fixes the bug, but I wanted to make sure this doesn't ring any alarm bells or follow-up work.


paulkaplan:
👍

mzgoddard:
Made an issue for the matter I mentioned. https://github.com/LLK/scratch-render/issues/208

PR #210, opened 2017-12-05
Fix mobile color bug

### LINKED ISSUE #168: incorrect color effect for white area on phone/pad
### Expected Behavior

It should work like in Desktop browser...

### Actual Behavior

When set color effect to something larger than zero (for example, 1...10...), the sprite's white area (maybe it is transparent area?) becomes black.

### Steps to Reproduce

On iPhone or Android:
1. Open llk.github.io/scratch-gui/develop
2. drag "set color effect to xxx" block in "looks" category to workspace
3. click the block to run it
4. issue as below:
img_3909

It works on my Mac OS X (MacBook Air) Safari browser, and my Ubuntu Chrome browser. But failed on both my iPhone (which is Safari) and my Android tablet (in Chrome browser). So, it seems not an issue of browser but an issue of "mobile" comparing with PC???

---
Moved from LLK/scratch-gui#674
/cc @gengshenghong

cwillisf:
### Resolves

Resolves #168

### Proposed Changes

This PR includes two changes:
- Adjust the value of `kEpsilon` for compatibility with mobile GPUs operating at a lower precision
- Simplify the `convertRGB2HSL` function and add comments explaining each step of the algorithm

### Reason for Changes

The `convertRGB2HSL` function does a few division operations where the denominator can sometimes be zero. In the formal definition these steps are defined piecewise, such as "S = { 0 if L=1, or C/(1-abs(2*L-1)) otherwise," which avoids the possibility of a division by zero. Since these cases always correspond to numerator of zero as well, though, we can avoid the conditional by adding a small "epsilon" value to the denominator. If the value is small enough it won't cause a significant difference in the output of the function, which will eventually (after adjustment by the color or brightness effect and conversion back to RGB) be quantized to 8 bits per channel anyway.

The problem with the old epsilon value is that on some mobile devices -- whether it's because they ignore the precision specification at the top of the shader, or perform certain operations at a lower precision, or maybe even just contain bugs -- the old value was too small to "count" as a non-zero denominator. This led to a divide-by-zero, resulting in `NaN`, which "infected" all components during the HSL-to-RGB conversion later. This would be rendered as black, when generally this case was triggered by white (L=1) as the input color.

I also simplified and commented the shader code for three reasons:
- The more of our code that can be understood by contributors and ATers, the better.
- I didn't fully understand the algorithm before, and I wanted to make it easier for myself and for others on the team to work on this code later.
- The new version may execute slightly faster, depending on details of the shader compiler (part of the video driver), but certainly shouldn't execute any slower.

### Test Coverage

We don't currently have any way to test visual output automatically, but I have tested this on a few mobile devices which previously exhibited the problem in #168 but now give correct results.

To test this yourself you can load project 24542731 in the Scratch 3.0 editor or `player.html` and press the green flag. Expected results:
- the orange areas of the cat should turn blue
- the black and white areas of the cat should remain unchanged

### DIFF
--- a/src/shaders/sprite.frag
+++ b/src/shaders/sprite.frag
@@ -43,24 +43,44 @@ varying vec2 v_texCoord;
 // Branchless color conversions based on code from:
 // http://www.chilliant.com/rgb2hsv.html by Ian Taylor
 // Based in part on work by Sam Hocevar and Emil Persson
+// See also: https://en.wikipedia.org/wiki/HSL_and_HSV#Formal_derivation
 
+// Smaller values can cause problems with "color" and "brightness" effects on some mobile devices
 const float kEpsilon = 1e-4;
 
-vec3 convertRGB2HCV(vec3 rgb)
-{
-	vec4 p = (rgb.g < rgb.b) ? vec4(rgb.bg, -1, 2.0/3.0) : vec4(rgb.gb, 0, -1.0/3.0);
-	vec4 q = (rgb.r < p.x) ? vec4(p.xyw, rgb.r) : vec4(rgb.r, p.yzx);
-	float c = q.x - min(q.w, q.y);
-	float h = abs((q.w - q.y) / (6.0 * c + kEpsilon) + q.z);
-	return vec3(h, c, q.x);
-}
-
+// Convert an RGB color to Hue, Saturation, and Lightness.
+// All components of input and output are expected to be in the [0,1] range.
 vec3 convertRGB2HSL(vec3 rgb)
 {
-	vec3 hcv = convertRGB2HCV(rgb);
-	float l = hcv.z - hcv.y * 0.5;
-	float s = hcv.y / (1.0 - abs(l * 2.0 - 1.0) + kEpsilon);
-	return vec3(hcv.x, s, l);
+	// Hue calculation has 3 cases, depending on which RGB component is largest, and one of those cases involves a "mod"
+	// operation. In order to avoid that "mod" we split the M==R case in two: one for G<B and one for B>G. The B>G case
+	// will be calculated in the negative and fed through abs() in the hue calculation at the end.
+	// See also: https://en.wikipedia.org/wiki/HSL_and_HSV#Hue_and_chroma
+	const vec4 hueOffsets = vec4(0.0, -1.0 / 3.0, 2.0 / 3.0, -1.0);
+
+	// temp1.xy = sort B & G (largest first)
+	// temp1.z = the hue offset we'll use if it turns out that R is the largest component (M==R)
+	// temp1.w = the hue offset we'll use if it turns out that R is not the largest component (M==G or M==B)
+	vec4 temp1 = rgb.b > rgb.g ? vec4(rgb.bg, hueOffsets.wz) : vec4(rgb.gb, hueOffsets.xy);
+
+	// temp2.x = the largest component of RGB ("M" / "Max")
+	// temp2.yw = the smaller components of RGB, ordered for the hue calculation (not necessarily sorted by magnitude!)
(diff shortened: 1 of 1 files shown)


paulkaplan:
Wow cool fix, and nice job annotating that code, that'll really be appreciated by someone in the future. 

Also led me down a rabbit hole of what the "k" prefix is used for (https://stackoverflow.com/questions/5016622/where-does-the-k-prefix-for-constants-come-from), so fwiw you probably don't need it since glsl has constants!

Did you find that new value by experimenting, or is that a documented precision problem (i.e. do you expect this fix to continue working, or need to be tweaked in the future?)

cwillisf:
Ha! You're right - I should probably remove the `k`. Call it an old habit :)

I did find the new value experimentally, but I don't think it'll be necessary to tweak it in the future. I'm reasonably confident that the problem was caused by mobile GPUs processing these values with 16-bit floats instead of 32-bit floats. I don't think 8-bit floats are allowed in GLSL so I think if we're safe on 16-bit floats we'll be safe on any platform which supports GLSL.

I looked into this in a bit more detail just now and actually learned something new! I had been thinking that saying `precision mediump float` would specify that your floats would always be 32-bit -- or more specifically, I thought `lowp` would always mean 16-bit, `mediump` would always mean 32-bit (like `float` in some typed languages), and `highp` would always mean 64-bit (like `double` in some typed languages). Surprise: this is not the case!

The precision of `lowp`, `mediump`, and `highp` are platform-specific, within some constraints specified by the spec. The spec specifies the meaning of `lowp`, `mediump`, and `highp` by range and precision, not by bits, but if you assume IEEE-style floating point formats then it works out to roughly:
- `lowp` must be at least 8 bits
- `mediump` must be at least 10 bits
- `highp` must be at least 16 bits

All current Mali GPUs use 16 bits for `mediump`, which is well within spec but definitely isn't 32 bits! That's why my phone, which uses a Mali chip, was vulnerable to #168.

Thanks for encouraging me to figure out the rest of the explanation! ;)

cwillisf:
An epsilon value of 1e-4 should work for all compliant implementations, given the `precision mediump float` at the top of this shader.

I verified this by using the handy calculator at https://oletus.github.io/float16-simulator.js/ with a 10-bit mantissa and 5-bit exponent. That's the smallest that either can be while still fulfilling the requirements for `mediump`. I then divided zero by 0.0001 and got a result of zero (good) -- not `NaN` (bad). Dividing by 0.00001 (1e-5) resulted in `NaN`, so 1e-4 seems good to me.

Bottom line: even though I can't seem to find any real GPUs that use a 15-bit float for `mediump` we'd be safe with `1e-4` on such a chip, as well as any chip with a larger `mediump`. :)

src/shaders/sprite.frag
```diff
@@ -43,24 +43,44 @@ varying vec2 v_texCoord;
...
-vec3 convertRGB2HCV(vec3 rgb)
-{
-	vec4 p = (rgb.g < rgb.b) ? vec4(rgb.bg, -1, 2.0/3.0) : vec4(rgb.gb, 0, -1.0/3.0);
-	vec4 q = (rgb.r < p.x) ? vec4(p.xyw, rgb.r) : vec4(rgb.r, p.yzx);
-	float c = q.x - min(q.w, q.y);
-	float h = abs((q.w - q.y) / (6.0 * c + kEpsilon) + q.z);
-	return vec3(h, c, q.x);
-}
+// Smaller values can cause problems with "color" and "brightness" effects on some mobile devices
+const float kEpsilon = 1e-4;
```
cwillisf: The comment here is meant as "you might think you can make this smaller but be aware that you might break some clients, especially ones running on mobile devices"

Issue #212, opened 2017-12-08
Touching color should work when sprites are not visible


paulkaplan:
From @towerofnix 

---
Expected (2.0) Behavior
"Touching color" returns true when the sprite overlaps the given color, regardless of whether or not the sprite is hidden. "Touching sprite" returns true if the sprite overlaps the other sprite, and the sprite is shown; if it is hidden, "touching sprite" returns false.

Actual Behavior
"Touching color" returns false when the sprite is hidden. "Touching sprite" behaves as described in 2.0.

Steps to Reproduce
Demo project: 2.0, 3.0. Move the sprite around using the mouse, and use the space bar to toggle the sprite between being shown and hidden. Observe the contents of the speech bubble while dragging.

(Note - in the speech bubble that shows up, you can see that, next to "touching (sprite)", sometimes "undefined" is being said instead of "false".. undefined is a falsey value, so this is okay, but maybe something to look into if it's not already reported somewhere..)

I don't know what the expected behavior is, but I would think that this incompatibility could be a source of issues in platformer projects?

Demo project 2: 2.0, 3.0. Notice how, in Scratch 3.0, when you hide the sprite, it falls through the floor! 🕳

Operating System and Browser
Firefox Nightly (59) on Debian Testing. I'm fairly certain this same bug was reported here.

paulkaplan:
Eek here are 4 cases I tested. 

|                 | Hidden       | Ghost 100 |
|-----------------|--------------|-----------|
| Touching sprite | not detected | detected  |
| Touching color  | detected     | detected  |

In scratch 2
![image](https://user-images.githubusercontent.com/654102/33773576-5fa9887a-dc06-11e7-9ba2-216aac6c163d.png)

In scratch 3
![image](https://user-images.githubusercontent.com/654102/33773718-d0575d2c-dc06-11e7-94d2-11d5021fc79e.png)

towerofnix:
Broken demo project links in the original post btw: demo 1 in [2.0](https://scratch.mit.edu/projects/191930440/#editor), [3.0](https://llk.github.io/scratch-gui/#191930440), demo 2 in [2.0](https://scratch.mit.edu/projects/191931857/#editor), [3.0](https://llk.github.io/scratch-gui/#191931857).

paulkaplan:
Cool thanks, verified with the falling cat project. 

[...]

PR #213, opened 2017-12-08
Ignore sprite visibility when testing for touching color

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-render/issues/212

### Proposed Changes

Make the draw step of the touching color test ignore sprite visibility. Use the same flag as `isStamping` but rename for more general use case.

Also return `false` when there are no candidate sprites to test nearby, instead of returning undefined.

### Reason for Changes

To bring the touching color behavior in line with scratch 2. 

### Test Coverage

Tested the demo project I made in the linked issue. Now it matches scratch 2

image

image

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1118,7 +1118,8 @@ class RenderWebGL extends EventEmitter {
             const drawable = this._allDrawables[drawableID];
             /** @todo check if drawable is inside the viewport before anything else */
 
-            // Hidden drawables (e.g., by a "hide" block) are not drawn unless stamping
+            // Hidden drawables (e.g., by a "hide" block) are not drawn unless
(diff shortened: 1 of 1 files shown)


paulkaplan:
@towerofnix thanks for the demo projects. I verified with the falling cat project and also added a commit to fix the `undefined` issue. That was a good catch, wouldn't have noticed because of the falsey-ness, but definitely should be `false`. 

sensing-report--2
sensing-report--3

src/RenderWebGL.js
```diff
@@ -1115,7 +1119,7 @@ class RenderWebGL extends EventEmitter {
             /** @todo check if drawable is inside the viewport before anything else */
 
             // Hidden drawables (e.g., by a "hide" block) are not drawn unless stamping
```
cwillisf: This comment is now only 90% accurate :)

PR #226, opened 2018-01-10
Check & report if browser supports this renderer

### LINKED ISSUE #220: Fail gracefully if WebGL isn't available
### Expected Behavior

Renderer should check whether WebGL is available and fail gracefully if it isn't

### Actual Behavior

Fatal errors in the console

### Steps to Reproduce

Run the GUI on a machine or browser that doesn't support WebGL, examples:
- old samsung chromebooks
- old macbook Air Jamaica

cwillisf:
### Resolves

Resolves #220 

### Proposed Changes

`RenderWebGL`'s constructor will now throw if TWGL init fails. There's also a new static method to check whether or not we expect TWGL init to succeed.

### Reason for Changes

This will prevent a flood of errors if the renderer thinks it's working but has no WebGL context. The new `isSupported` method can be used by client code (such as in `scratch-gui`) to provide feedback to the user outside of renderer-construction time. 

### Test Coverage

Because this code depends on the browser so heavily it's not possible to test it automatically on our current test framework. I did test it in a few situations, with and without WebGL, WebGL 1 and 2, and it seems to do the right thing.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -59,6 +59,31 @@ const FENCE_WIDTH = 15;
 
 class RenderWebGL extends EventEmitter {
     /**
+     * Check if this environment appears to support this renderer before attempting to create an instance.
+     * Catching an exception from the constructor is also a valid way to test for (lack of) support.
+     * @param {canvas} [optCanvas] - An optional canvas to use for the test. Otherwise a temporary canvas will be used.
+     * @returns {boolean} - True if this environment appears to support this renderer, false otherwise.
+     */
+    static isSupported (optCanvas) {
+        try {
+            // Create the context the same way that the constructor will: attributes may make the difference.
+            return !!RenderWebGL._getContext(optCanvas || document.createElement('canvas'));
+        } catch (e) {
+            return false;
+        }
+    }
+
+    /**
+     * Ask TWGL to create a rendering context with the attributes used by this renderer.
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -58,6 +58,31 @@ const FENCE_WIDTH = 15;
...
+    }
+
+    /**
+     * Ask TWGL to create a rendering context with the attributes used by this renderer.
+     * @param {canvas} canvas - attach the context to this canvas.
+     * @returns {WebGLRenderingContext} - a TWGL rendering context (backed by either WebGL 1.0 or 2.0).
+     * @private
+     */
+    static _getContext (canvas) {
+        return twgl.getContext(canvas, {alpha: false, stencil: true});
```
paulkaplan: @cwillisf I notice that the twgl method here changed from `getWebGLContext` to `getContext`. Is there an important reason for that?

paulkaplan: Oops just looked at the docs and saw that this will create a webgl2 context if possible. I'm not sure we want to introduce that change here. If we want to support webgl 2, it might be a good idea to do that as a self contained unit so that we can see if there are any compatibility issues or things we can take advantage of.

cwillisf: Hmm... that's fair. I'll change this back & file a lower-priority issue.

cwillisf:
The idea of changing from `getWebGLContext` to `getContext` is now tracked in #230

PR #231, opened 2018-01-12
Extract svg-renderer out into a separate module

cwillisf:
### Proposed Changes

The SVG rendering and "fixup" code is now in LLK/scratch-svg-renderer; this change removes the now-duplicate from the `scratch-render` repository.

### Reason for Changes

The SVG code will be used in more repositories than just this one. Ideally, SVG fixup will be done at load-time and only for legacy SVGs.

This change also supports WIP for LLK/scratch-paint#206 and LLK/scratch-gui#762

### Test Coverage

Since SVG features depend on browser features, we can't test them with the current automated test framework in this repository. However, the existing tests still pass.


package.json
```diff
@@ -41,7 +41,7 @@
     "json": "^9.0.4",
     "linebreak": "0.3.0",
     "raw-loader": "^0.5.1",
-    "scratch-render-fonts": "git+https://github.com/LLK/scratch-render-fonts.git",
+    "scratch-svg-renderer": "git+https://github.com/LLK/scratch-svg-renderer.git",
```
fsih: I think this is okay for now, but we should probably have scratch-svg-renderer deploy to npm like the other repos eventually.

src/SVGSkin.js
```diff
@@ -1,9 +1,9 @@
 const twgl = require('twgl.js');
 
+const Silhouette = require('./Silhouette');
 const Skin = require('./Skin');
-const SvgRenderer = require('./svg-quirks-mode/svg-renderer');
 
-const Silhouette = require('./Silhouette');
+import SvgRenderer from 'scratch-svg-renderer';
```
paulkaplan: I think we should stick with using require if we are using it elsewhere and not switch to import?

cwillisf: 👍 the `scratch-svg-renderer` was using `export default` so I used `import` here (otherwise the build would fail). However, I accidentally changed it to `module.exports = ...` when I merged from `scratch-render` into `scratch-svg-renderer` so... I guess I'll change these too ;)

package.json
cwillisf: Tracking this in LLK/scratch-svg-renderer#1

src/util/svg-text-bubble.js
```diff
@@ -1,5 +1,5 @@
 const SVGTextWrapper = require('./svg-text-wrapper');
-const SVGRenderer = require('../svg-quirks-mode/svg-renderer');
+const SvgRenderer = require('scratch-svg-renderer');
```
paulkaplan: The capitalization change here is going to break the usage later in the file.

cwillisf: To you: good catch! Thanks!
To self: BAH! Watch those copy & pastes there, bucko!

cwillisf:
Postponing this until after preview

PR #232, opened 2018-01-17
Re-rasterize SVGs when scaled up

### LINKED ISSUE #24: SVG re-rasterization for varying sizes
We've talked about this at some point. Currently, when you scale up a sprite, it gets blurry. I think this is because the SVG rasterization only happens once, while it needs to happen either on resize, or generate multiple sizes at the beginning.

cwillisf:
### Resolves

Resolves #24

### Proposed Changes

When an SVG costume is scaled up sufficiently, re-rasterize it to a new texture with a higher resolution, within limits. The maximum scale factor is the largest power of two which scales the width and height of the SVG without exceeding a predefined limit. The limit is currently set to 2048, which matches the limit used by Scratch 2.0.

### Reason for Changes

Vector costumes have theoretically infinite resolution, but need to be rasterized to "bitmap" images for use with WebGL. Rasterizing too large wastes video texture memory and makes the video hardware work harder, but rasterizing too small makes the costume look bad when scaled up. This change makes the renderer re-rasterize a particular costume on demand based on the scale used to display it.

### Test Coverage

See #24 for manual test cases. Here are the results of this change, but note that project 113396270 is limited by the resolution of the stamping layer.

</img>

image

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -88,7 +88,6 @@ class SVGSkin extends Skin {
             });
         }
 
-        /** @todo re-render a scaled version if the requested scale is significantly larger than the current render */
         return this._texture;
     }
 
@@ -103,7 +102,6 @@ class SVGSkin extends Skin {
         this._svgRenderer.fromString(svgData, 1, () => {
             const gl = this._renderer.gl;
             this._textureScale = this._maxTextureScale = 1;
-            this._svgRenderer.extraScale = 1;
             if (this._texture) {
                 gl.bindTexture(gl.TEXTURE_2D, this._texture);
                 gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this._svgRenderer.canvas);
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -5,6 +5,8 @@ const SvgRenderer = require('./svg-quirks-mode/svg-renderer');
 
 const Silhouette = require('./Silhouette');
 
+const MAX_TEXTURE_DIMENSION = 2048;
+
 class SVGSkin extends Skin {
     /**
      * Create a new SVG skin.
@@ -24,6 +26,12 @@ class SVGSkin extends Skin {
 
         /** @type {WebGLTexture} */
         this._texture = null;
+
+        /** @type {number} */
+        this._textureScale = 1;
+
+        /** @type {Number} */
+        this._maxTextureScale = 0;
 
         this._silhouette = new Silhouette();
     }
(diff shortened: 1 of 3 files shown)


cwillisf:
CC @mzgoddard

src/SVGSkin.js
```diff
@@ -25,6 +27,12 @@ class SVGSkin extends Skin {
         /** @type {WebGLTexture} */
         this._texture = null;
 
+        /** @type {number} */
+        this._textureScale = 1;
+
+        /** @type {Number} */
+        this._maxTextureScale = 0;
```
paulkaplan: Since this is reset to the same value as _textureScale (i.e. 1) in `setSVG`, should this be initialized to 1 here also?

src/SVGSkin.js
```diff
@@ -57,11 +65,29 @@ class SVGSkin extends Skin {
...
+            this._svgRenderer._draw(this._textureScale, () => {
+                if (this._textureScale === newScale) {
+                    const gl = this._renderer.gl;
+                    gl.bindTexture(gl.TEXTURE_2D, this._texture);
+                    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this._svgRenderer.canvas);
+                }
+            });
+        }
+
         /** @todo re-render a scaled version if the requested scale is significantly larger than the current render */
```
paulkaplan: Can remove this `@todo` comment!

src/SVGSkin.js
```diff
@@ -74,8 +100,10 @@ class SVGSkin extends Skin {
      * @fires Skin.event:WasAltered
      */
     setSVG (svgData, rotationCenter) {
-        this._svgRenderer.fromString(svgData, () => {
+        this._svgRenderer.fromString(svgData, 1, () => {
             const gl = this._renderer.gl;
+            this._textureScale = this._maxTextureScale = 1;
+            this._svgRenderer.extraScale = 1;
```
paulkaplan: This does't seem to be used anywhere?

src/SVGSkin.js
```diff
@@ -57,11 +65,29 @@ class SVGSkin extends Skin {
...
+        while ((newScale < this._maxTextureScale) && (requestedScale >= 1.5 * newScale)) {
+            newScale *= 2;
+        }
+        if (this._textureScale !== newScale) {
+            this._textureScale = newScale;
+            this._svgRenderer._draw(this._textureScale, () => {
+                if (this._textureScale === newScale) {
+                    const gl = this._renderer.gl;
+                    gl.bindTexture(gl.TEXTURE_2D, this._texture);
+                    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this._svgRenderer.canvas);
```
paulkaplan: I noticed that other places where we are updating the texture, we also update the silhouette. I assume that isn't needed here?

src/SVGSkin.js
```diff
@@ -91,6 +119,13 @@ class SVGSkin extends Skin {
                 this._texture = twgl.createTexture(gl, textureOptions);
                 this._silhouette.update(this._svgRenderer.canvas);
             }
+
+            const maxDimension = Math.max(this._svgRenderer.canvas.width, this._svgRenderer.canvas.height);
+            let testScale = 2;
+            for (testScale; maxDimension * testScale <= MAX_TEXTURE_DIMENSION; testScale *= 2) {
+                this._maxTextureScale = testScale;
```
paulkaplan: This block might benefit from a comment, at first it wasn't clear what was going on here. Is my understanding right: it looks like we are precomputing the max texture scale after rendering the texture the first time, when it was rendered presumably at a scale of 1. This was surprising to me at first because we have a very similar loop in `getTexture`, but it is running on scale instead of dimensions.

paulkaplan:
@cwillisf I pushed a commit to this branch addressing my comments including the caching of the image element.

src/SVGSkin.js
cwillisf: Yeah, I'm not very happy with the way this is set up. I can certainly add a comment but I'd like to find a way to simplify this code now that the basics are working.

PR #263, opened 2018-04-24
Optimize isTouchingDrawables to use Silhouette

gnarf:
This is a work in progress and has a lot of debugging stuff still going on, please feel free to comment on any of it (including the debugging, it can only help me remember to remove it later! <3)

### Proposed Changes

* Take the `isTouchingDrawables` off the GPU

### Reason for Changes

* `gl.readPixels` is the ultimate in slow

### Test Coverage

* See #265

### DIFF
--- a/src/BitmapSkin.js
+++ b/src/BitmapSkin.js
@@ -1,7 +1,6 @@
 const twgl = require('twgl.js');
 
 const Skin = require('./Skin');
-const Silhouette = require('./Silhouette');
 
 class BitmapSkin extends Skin {
     /**
@@ -24,8 +23,6 @@ class BitmapSkin extends Skin {
 
         /** @type {Array<int>} */
         this._textureSize = [0, 0];
-
-        this._silhouette = new Silhouette();
     }
 
     /**
@@ -117,14 +114,6 @@ class BitmapSkin extends Skin {
         return [bitmapData.width, bitmapData.height];
     }
 
-    /**
-     * Does this point touch an opaque or translucent point on this skin?
-     * @param {twgl.v3} vec A texture coordinate.
-     * @return {boolean} Did it touch?
-     */
-    isTouching (vec) {
-        return this._silhouette.isTouching(vec);
-    }
 }
 
 module.exports = BitmapSkin;
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -376,7 +376,7 @@ class Drawable {
      * @return {boolean} True if the world position touches the skin.
      */
     isTouching (vec) {
-        if (!this.skin) {
+        if (!(this.skin && this._visible)) {
             return false;
         }
 
@@ -411,7 +411,32 @@ class Drawable {
         // Apply texture effect transform.
         EffectTransform.transformPoint(this, localPosition, localPosition);
 
-        return this.skin.isTouching(localPosition);
+        if (this.useNearest) {
+            return this.skin.isTouchingNearest(localPosition);
+        }
+        return this.skin.isTouchingLinear(localPosition);
+    }
+
+    /**
+     * Should the drawable use NEAREST NEIGHBOR or LINEAR INTERPOLATION mode
+     */
+    get useNearest () {
+        // We can't use nearest neighbor unless we are a multiple of 90 rotation
+        if (this._direction % 90 !== 0) {
+            return false;
+        }
+
+        // Raster skins (bitmaps) should always prefer nearest neighbor
+        if (this.skin.isRaster) {
+            return true;
+        }
+
+        // If the scale of the skin is very close to 100 (0.99999 variance is okay I guess)
+        if (Math.abs(this.scale[0]) > 99 && Math.abs(this.scale[0]) < 101 &&
+            Math.abs(this.scale[1]) > 99 && Math.abs(this.scale[1]) < 101) {
+            return true;
+        }
+        return false;
     }
 
     /**
--- a/src/EffectTransform.js
+++ b/src/EffectTransform.js
@@ -26,10 +26,9 @@ class EffectTransform {
      * @param {Drawable} drawable The drawable whose effects to emulate.
      * @param {twgl.v3} vec The texture coordinate to transform.
      * @param {?twgl.v3} dst A place to store the output coordinate.
-     * @return {twgl.v3} The coordinate after being transform by effects.
+     * @return {twgl.v3} dst - The coordinate after being transform by effects.
      */
-    static transformPoint (drawable, vec, dst) {
-        dst = dst || twgl.v3.create();
+    static transformPoint (drawable, vec, dst = twgl.v3.create()) {
         twgl.v3.copy(vec, dst);
 
         const uniforms = drawable.getUniforms();
--- a/src/Rectangle.js
+++ b/src/Rectangle.js
@@ -35,6 +35,7 @@ class Rectangle {
         this.right = -Infinity;
         this.top = -Infinity;
         this.bottom = Infinity;
+
         for (let i = 0; i < points.length; i++) {
             const x = points[i][0];
             const y = points[i][1];
@@ -115,6 +116,41 @@ class Rectangle {
     }
 
     /**
+     * Compute the intersection of two bounding Rectangles.
+     * Could be an impossible box if they don't intersect.
+     * @param {Rectangle} a One rectangle
+     * @param {Rectangle} b Other rectangle
+     * @param {?Rectangle} result A resulting storage rectangle  (safe to pass
+     *                            a or b if you want to overwrite one)
+     * @returns {Rectangle} resulting rectangle
+     */
+    static intersect (a, b, result = new Rectangle()) {
+        result.left = Math.max(a.left, b.left);
+        result.right = Math.min(a.right, b.right);
+        result.top = Math.min(a.top, b.top);
+        result.bottom = Math.max(a.bottom, b.bottom);
+
+        return result;
+    }
+
+    /**
+     * Compute the union of two bounding Rectangles.
+     * @param {Rectangle} a One rectangle
+     * @param {Rectangle} b Other rectangle
+     * @param {?Rectangle} result A resulting storage rectangle  (safe to pass
+     *                            a or b if you want to overwrite one)
+     * @returns {Rectangle} resulting rectangle
+     */
+    static union (a, b, result = new Rectangle()) {
+        result.left = Math.min(a.left, b.left);
+        result.right = Math.max(a.right, b.right);
+        // Scratch Space - +y is up
+        result.top = Math.max(a.top, b.top);
+        result.bottom = Math.min(a.bottom, b.bottom);
+        return result;
+    }
+
+    /**
      * Width of the Rectangle.
      * @return {number} Width of rectangle.
      */
@@ -129,6 +165,7 @@ class Rectangle {
     get height () {
         return Math.abs(this.top - this.bottom);
     }
+
 }
 
 module.exports = Rectangle;
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -13,6 +13,9 @@ const SVGSkin = require('./SVGSkin');
 const SVGTextBubble = require('./util/svg-text-bubble');
 const EffectTransform = require('./EffectTransform');
 
+const __isTouchingDrawablesPoint = twgl.v3.create();
+const __candidatesBounds = new Rectangle();
+
 /**
  * @callback RenderWebGL#idFilterFunc
  * @param {int} drawableID The ID to filter.
@@ -515,14 +518,14 @@ class RenderWebGL extends EventEmitter {
         const gl = this._gl;
         twgl.bindFramebufferInfo(gl, this._queryBufferInfo);
 
-        const bounds = this._touchingBounds(drawableID);
-        if (!bounds) {
+        const candidates = this._candidatesTouching(drawableID, this._drawList);
+        if (candidates.length === 0) {
             return false;
         }
-        const candidateIDs = this._filterCandidatesTouching(drawableID, this._drawList, bounds);
-        if (!candidateIDs) {
-            return false;
-        }
+
+        const bounds = this._candidatesBounds(candidates);
+
+        const candidateIDs = candidates.map(({id}) => id);
 
         // Limit size of viewport to the bounds around the target Drawable,
         // and create the projection matrix for the draw.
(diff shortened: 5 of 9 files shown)


src/Rectangle.js
```diff
@@ -129,6 +129,32 @@ class Rectangle {
     get height () {
         return Math.abs(this.top - this.bottom);
     }
+
+    points (vec) {
```
gnarf: Will add comments here...

Basically I wrote this without using generator functions because `regeneratorRuntime` is kinda a big dependency.

src/RenderWebGL.js
```diff
@@ -615,9 +621,30 @@ class RenderWebGL extends EventEmitter {
...
         }
 
+        measure.start('newWay');
+
+        const point = twgl.v3.create(0, 0);
+        const iterator = bounds.points(point);
+        const drawable = this._allDrawables[drawableID];
+        const candidates = candidateIDs.map(id => this._allDrawables[id]);
+        let newWay = false;
+        while (iterator.next().value) {
```
gnarf: yes, looping over EVERY pixel and asking the Silhouette is still much faster than reading those pixels back from the GPU

gnarf: If anyone has any thoughts on an even more performant approach - please chime in!

src/RenderWebGL.js
```diff
@@ -615,9 +621,30 @@ class RenderWebGL extends EventEmitter {
...
+
+        measure.end('newWay');
+        measure.start('oldWay');
+
+        // return newWay;
+
         // Limit size of viewport to the bounds around the target Drawable,
         // and create the projection matrix for the draw.
-        gl.viewport(0, 0, bounds.width, bounds.height);
+        gl.viewport(0, 0, bounds.width + 1, bounds.height + 1);
```
gnarf: I was actually detecting a positive here that the old way didn't -- the old way doesn't detect collisions on the far right side of a cat whisker for instance (when the cat is asking...)

src/RenderWebGL.js
```diff
@@ -629,30 +656,46 @@ class RenderWebGL extends EventEmitter {
             gl.stencilFunc(gl.ALWAYS, 1, 1);
             gl.stencilOp(gl.KEEP, gl.KEEP, gl.REPLACE);
             gl.colorMask(false, false, false, false);
-            this._drawThese([drawableID], ShaderManager.DRAW_MODE.silhouette, projection);
+            this._drawThese([drawableID, ...candidateIDs], ShaderManager.DRAW_MODE.silhouette, projection);
```
gnarf: also done for debugging purposes -- to benchmark I removed this change

src/RenderWebGL.js
```diff
@@ -629,30 +656,46 @@ class RenderWebGL extends EventEmitter {
...
 
             gl.stencilFunc(gl.EQUAL, 1, 1);
             gl.stencilOp(gl.KEEP, gl.KEEP, gl.KEEP);
             gl.colorMask(true, true, true, true);
 
-            this._drawThese(candidateIDs, ShaderManager.DRAW_MODE.silhouette, projection,
-                {idFilterFunc: testID => testID !== drawableID}
-            );
+            this._drawThese(candidateIDs.concat([drawableID]), ShaderManager.DRAW_MODE.silhouette, projection);
```
gnarf: remove `concat` here for a truer bench of the "old way" - but this all gets blown away long term

src/RenderWebGL.js
```diff
@@ -629,30 +656,46 @@ class RenderWebGL extends EventEmitter {
...
         } finally {
             gl.colorMask(true, true, true, true);
             gl.disable(gl.STENCIL_TEST);
         }
 
-        const pixels = new Uint8Array(Math.floor(bounds.width * bounds.height * 4));
-        gl.readPixels(0, 0, bounds.width, bounds.height, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+        const pixels = new Uint8Array(Math.floor((bounds.width+1) * (bounds.height+1) * 4));
+        gl.readPixels(0, 0, bounds.width+1, bounds.height+1, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
```
gnarf: again -- bound rectangles, can we talk about this one?  Is it just me or does a box from 10-20 have 11 pixels width? assuming 10 and 20 are output

src/RenderWebGL.js
```diff
@@ -600,7 +606,7 @@ class RenderWebGL extends EventEmitter {
      * @returns {boolean} True iff the Drawable is touching one of candidateIDs.
      */
     isTouchingDrawables (drawableID, candidateIDs) {
-        candidateIDs = candidateIDs || this._drawList;
+        candidateIDs = (candidateIDs || this._drawList).filter(id => id !== drawableID);
```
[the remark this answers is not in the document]
gnarf: I actually realized that the `_filterCandidates` or whateve that function is called did this, so I undid this...

src/Silhouette.js
```diff
@@ -53,18 +56,39 @@ class Silhouette {
...
+        const px = x % 1;
+        const py = y % 1;
+        const val = (this.point(xf, yf) * px * py) +
+            (this.point(xf + 1, yf) * (1 - px) * py) +
+            (this.point(xf, yf + 1) * px * (1 - py)) +
+            (this.point(xf + 1, yf + 1) * (1 - px) * (1 - py));
+        if (Math.floor(val) >= 4) {
+            return val;
+        }
+        return false;
```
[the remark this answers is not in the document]
gnarf: yeah this was a debug thing I think -- was trying to see what the values were for some low numbers, thanks for catching that one :)

gnarf:
You're totally write about the GC implications of the generator pattern, I was hoping for the `break` functionality to be a bit more obvious, but I think I could write a `Rectangle.somePixel(iterator)` that would be cleaner / less GC

gnarf:
> Just to be clear: have you benchmarked much on mobile?

nope - but I only have a really high end device available at my house right now (I need to pick up some lower end test devices from the office soon) - if you have a setup that would let you do some of this bench, I'm into knowing how it fares :)

cwillisf:
OK, so: bounding boxes.

Suppose we have a 4x4 pixel stage, and on it we center a sprite which is 2 pixels wide and 3 pixels tall:
image

At render time, the pixels in columns 0 and 3 will be white and the pixels in (1,1), (2,1), (1,2), and (2,2) will be blue. The pixels in (1,0), (2,0), (1,3), and (2,3) will theoretically be half-blue-half-white; the actual display depends on what filtering mode is active. (Fun fact: OpenGL and DirectX have different rules regarding edge rounding in cases like this.)

In stage coordinates, the bounds of the sprite would be `{x:1, y:0.5, w:2, h:3}` or `{left:1, right:2, top:0.5, bottom:3.5}` depending on your notation. Those `.5`s are the tricky part: if you want to clip an integer-pixel section of the stage which minimally includes every pixel the sprite could affect, you need to clip `{x:1, y:0, w:2, h:4}` or `{left:1, right:2, top:0, bottom:3}`! Either way, those bounds make it look like the sprite is 1 pixel taller than it actually is, but maintain the expected width.

To be safe, I believe the correct behavior is: `{left: floor(x), right: ceil(x+width), top: floor(y), bottom: ceil(y+height)}`. That should handle cases where we center on a pixel as well as cases where we center between pixels.

gnarf:
> To be safe, I believe the correct behavior is: {left: floor(x), right: ceil(x+width), top: floor(y), bottom: ceil(y+height)}. That should handle cases where we center on a pixel as well as cases where we center between pixels.

It's also hard to understand some times, but I think top is `ceil(y)` and bottom is `floor(y-height)` -- I'm pretty sure most of these rectangles are in "scratch coordinates"

cwillisf:
Ah - right. Sorry, I was thinking in generic terms -- you're totally right about flipping the Y axis in Scratch :)

src/Drawable.js
```diff
@@ -376,7 +376,7 @@ class Drawable {
      * @return {boolean} True if the world position touches the skin.
      */
     isTouching (vec) {
-        if (!this.skin) {
+        if (!(this.skin && this._visible)) {
```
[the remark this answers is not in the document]
gnarf: sounds like a unit/integration test to me ;)

src/RenderWebGL.js
```diff
@@ -661,11 +689,27 @@ class RenderWebGL extends EventEmitter {
                 pixels[pixelBase + 1],
                 pixels[pixelBase + 2]);
             if (pixelID > RenderConstants.ID_NONE) {
-                return true;
+                if (newWay !== true) {
```
gnarf: also remember to comment out console logs if you want to measure

gnarf:
Current "known issue" -- transparent fill boxes don't collide, but only when rotated off 90???

image

image

src/BitmapSkin.js
```diff
@@ -122,8 +122,8 @@ class BitmapSkin extends Skin {
      * @param {twgl.v3} vec A texture coordinate.
      * @return {boolean} Did it touch?
      */
-    isTouching (vec) {
-        return this._silhouette.isTouching(vec);
+    isTouching (vec, interpolate) {
```
mzgoddard: I don't know about this API ...

At minimum, update the jsdoc block.

src/EffectTransform.js
```diff
@@ -26,10 +26,9 @@ class EffectTransform {
      * @param {Drawable} drawable The drawable whose effects to emulate.
      * @param {twgl.v3} vec The texture coordinate to transform.
      * @param {?twgl.v3} dst A place to store the output coordinate.
-     * @return {twgl.v3} The coordinate after being transform by effects.
+     * @return {twgl.v3} dst - The coordinate after being transform by effects.
      */
-    static transformPoint (drawable, vec, dst) {
-        dst = dst || twgl.v3.create();
+    static transformPoint (drawable, vec, dst = twgl.v3.create()) {
```
mzgoddard: :heart:

src/Rectangle.js
```diff
@@ -129,6 +129,35 @@ class Rectangle {
...
+    /**
+     * Will pass a vector for each integer point within the bounds to a test function.
+     * If any testFunction returns truthy, the iteration will stop and return true.
+     * Otherwise, this will return false.
+     *
+     * @param {function} testFunction(point) should return true if iteration should stop.
+     * @param {?twgl.v3|array<number>} vec Storage for a vector to use for the iteration.
+     * @return {boolean} true if any testFunction returned true.
+     */
+    somePoint (testFunction, vec = [0, 0, 0]) {
```
mzgoddard: `twgl.v3.create()`

mzgoddard: I think more canonical names (in math libraries) for this method might be `overlapsPoint` or `intersectsPoint`?

src/Silhouette.js
```diff
@@ -53,18 +56,36 @@ class Silhouette {
...
      * @return {boolean} Did the point touch?
      */
-    isTouching (vec) {
-        const x = Math.floor(vec[0] * this._width);
-        const y = Math.floor(vec[1] * this._height);
-        return (
-            x < this._width && x >= 0 &&
-            y < this._height && y >= 0 &&
-            this._data[(y * this._width) + x] !== 0);
+    isTouching (vec, interpolate) {
```
mzgoddard: Is nearest neighbor `isTouching` not equivalent to interpolated `isTouching`?

src/BitmapSkin.js
gnarf: yeah - maybe we should always interp?

src/Rectangle.js
gnarf: it's an iterator?  it's basically `points().some()` so I called it `.somePoint()`

I also didn't `twgl.v3.create` here because rectangle didn't already import twgl - and I don't think it cares what you use as an iterator.

src/Silhouette.js
gnarf: it's quite a bit different if you disable interpolates - it looks much worse on the fuzzy edges of especially rotated elements

mzgoddard: Yeah I grokked this morning how they are different.

src/Rectangle.js
mzgoddard: Hmm. Rather I suppose the iterator would be a `iteratePoints` member. The rectangle isn't defined as a collection of points but a bounding shape on a plane which can have points inside. Those can be infinite though so you need to define parameters on the set of points you want and `Symbol.iterator` will not have that option.

src/BitmapSkin.js
mzgoddard: Uh. I'm not sure about the default, interpolation is fine as the default.

I do think we should have two separate `isTouching` methods with descriptive names whether they use nearest neighbor or interpolation. We could keep `isTouching` as a third method but it would only have the current one argument.

gnarf:
Had @thisandagain  run a test on iPad:

```
ascii [2:27 PM]
*iPad Mini 4 running iOS 11.3*
Current: 184 steps looped | 3496 executions
New: 703 steps looped | 13357 executions
```

So about 4x faster on the iPad mini -- I'll take it --- same numbers on  my desktop give me about 9-10x more executions

src/RenderWebGL.js
```diff
@@ -940,27 +907,47 @@ class RenderWebGL extends EventEmitter {
      * could possibly intersect the given bounds.
      * @param {int} drawableID - ID for drawable of query.
      * @param {Array<int>} candidateIDs - Candidates for touching query.
-     * @param {Rectangle} bounds - Bounds to limit candidates to.
-     * @return {?Array<int>} Filtered candidateIDs, or null if none.
+     * @return {?Array< {id, drawable, intersection} >} Filtered candidates with useful data.
```
gnarf: I wasn't sure how to best annotate this, anyone got a help?

gnarf:
@cwillisf -- I had @mzgoddard show me the benchmarking suite for this stuff - though I had to write a new project to test collisions hard - https://llk.github.io/scratch-gui/#219313833 

I've done a bunch of perf testing on browserstack's "Real Device Cloud" going from like Galaxy Tab 10.4/Nexus 7 to iPad pros and newer android tablets also.

On the worst devices, we see a 2x-3x improvement in number of collision detect operations performed, on the best we see 9x-10x improvements.  Most of the 3-4 year old devices land around 3-4x improvement.

NONE were slower using this branch

src/RenderWebGL.js
cwillisf: I usually end up using `@typedef` in this kind of situation. A good example to check out is `extension-support/define-messages.js` in scratch-vm. That might be overkill for this situation, but it's certainly an option.

PR #265, opened 2018-04-24
Quick example of a collision test integration test

gnarf:
(wanted to post this, sorry to not edit template yet, but want a travis run for testing)

### Resolves

### Proposed Changes

### Reason for Changes

### Test Coverage


gnarf:
> Are all the .travis.yml settings necessary? In particular, do you need dist and sudo?

sudo might not be -- dist is though -- they dont have chrome stable on other dists according to travis docs --- i'll do a squash rebase eventually to clean up i think

gnarf:
```
✓ gnarf@pro:scratch-render (vm-integration-tests=)$ tap test/integration -R tap
TAP version 13
# Subtest: test/integration/scratch-tests.js
    # Subtest: cat-touches-box.sb2
        1..2
        ok 1 - cat touches box
        ok 2 - box touches cat
    ok 1 - cat-touches-box.sb2 # time=2354.809ms
    
    # Subtest: tippy-toe-collision.sb2
        1..2
        ok 1 - cat touches box
        ok 2 - box touches cat
    ok 2 - tippy-toe-collision.sb2 # time=852.169ms
    
    1..2
    # time=3501.541ms
ok 1 - test/integration/scratch-tests.js # time=3885.627ms

1..1
# time=3896.699ms
```

gnarf:
@cwillisf rebased for a cleaner history -- "how about now" should mostly be "done"

package.json
```diff
@@ -18,8 +18,8 @@
     "prepublish": "npm run build",
     "prepublish-watch": "npm run watch",
     "start": "webpack-dev-server",
-    "tap": "./node_modules/.bin/tap ./test/unit/*.js",
-    "test": "npm run lint && npm run docs && npm run tap",
+    "tap": "./node_modules/.bin/tap ./test/unit/*.js test/integration",
```
kchadha: here we have `./test` ... followed by `test` is that intentional? The two should be consistent.

package.json
```diff
@@ -18,8 +18,8 @@
...
+    "test": "npm run lint && npm run docs && npm run build && npm run tap",
```
kchadha: why is the build step in here? This is inconsistent with the use of the test script in the other repos.

test/integration/index.html
```diff
@@ -0,0 +1,34 @@
+<body>
+    <script src="../../node_modules/scratch-vm/dist/web/scratch-vm.js"></script>
+    <script src="../../node_modules/scratch-storage/dist/web/scratch-storage.js"></script>
+    <script src="../../dist/web/scratch-render.js"></script>
+    <canvas id="test" width="480" height="360"></canvas>
+    <input type="file" id="file" name="file">
```
kchadha: This input doesn't have a closing tag

test/integration/scratch-tests.js
```diff
@@ -0,0 +1,92 @@
+/* global vm, Promise */
+const {Chromeless} = require('chromeless');
+const test = require('tap').test;
+const path = require('path');
+const fs = require('fs');
+const chromeless = new Chromeless();
+
+const testFile = (file) => {
```
kchadha: Unnecessary parens around single argument in arrow function.

test/integration/scratch-tests.js
```diff
@@ -0,0 +1,92 @@
...
+            return reporters.comment(text);
+        };
+
+        result.forEach(reportVmResult);
+        if (!didPlan) {
+            t.comment('did not say "plan NUMBER_OF_TESTS"');
+        }
+
+        // End must be called so that tap knows the test is done. If
+        // the test has an SAY "end" block but that block did not
```
kchadha: `test has an` -> `test has a`

package.json
gnarf: hrm -- so they are slightly different - also this doesn't need to be this complex...

`tap test/unit test/integration` works just fine in an npm command the `./node_modules/.bin` is in path for those commands... I'll update this.

package.json
gnarf: In order to do integration tests, we need to make sure the `dist/web/scratch-render.js` is up-to-date, I feel like anything doing integration tests likely does a build before the tests?

test/integration/index.html
gnarf: `<input>` is a "void element" and requires no closing tag (in fact adding one can be bad) -- I could do the `<input />` thing but its also not nessecary

http://www.456bereastreet.com/archive/201005/void_empty_elements_and_self-closing_start_tags_in_html/

test/integration/scratch-tests.js
gnarf: if this is a rule we like to follow, can we enable `"arrow-parens": ["error", "as-needed"]` in the eslint rules?

gnarf: it seems this rule is enabled in the `scratch/es6` ruleset which isn't used in this repo - enabling that...

gnarf:
@kchadha think I got all your nits nitted

package.json
kchadha: Yeah, but usually not in the `test` script itself, e.g. see `scratch-vm`

gnarf:
> IMO it'd be best to squash-merge this when the time comes, just because there are so many "how about now?" commits (as there always are when dealing with Travis...)

It's probably best to squash merge again, I'm a noisy committer and usually prefer my branches squash before landing anyway :)

test/integration/scratch-tests.js
```diff
@@ -0,0 +1,113 @@
...
+            // This function is run INSIDE the integration chrome browser via some
+            // injection and .toString() magic.  We can return some "simple data"
+            // back across as a promise, so we will just log all the says that happen
+            // for parsing after.
+
+            // this becomes the `says` in the outer scope
+            const messages = [];
+            const TIMEOUT = 5000;
+
+            vm.runtime.on('SAY', (_, __, message) => {
```
mzgoddard: What if in the index.html these were stored in a `<pre><code>` wrapper? After the threads complete, we'd then read out that element.

gnarf: I think it's a bit specific - I had imagined we might make other types of integration tests using the same index html/pattern..

I.E. a "click to pick" is a totally different valid integration test here if we could figure out how to script it.

test/integration/scratch-tests.js
```diff
@@ -0,0 +1,113 @@
...
+
+                        await new Promise(resolve => setTimeout(resolve, 50));
+                    }
+
+                    return messages;
+                });
+        });
+
+        // Map string messages to tap reporting methods. This will be used
+        // with events from scratch's runtime emitted on block instructions.
```
cwillisf: Looks like these comments should be un-indented one notch

test/integration/scratch-tests.js
cwillisf: I'm personally not very picky about the details of the message collection approach. I always imagined that we'd implement a specific block to emit testing data, but seeing this I'm not sure what the advantage would be if we're crafting the Scratch project specifically for the test.

That said, it might be helpful to have a `<pre><code>` display for the purposes of developing & debugging tests.

gnarf: I don't want to refactor it quite yet for say logging in the index.html yet -- we could add all sorts of debugging tools there I would think, but trying to keep it as simple as I can to ensure the test isn't flaky because of environment.  In other words, let's land without changing this.

test/integration/scratch-tests.js
gnarf: fixed

Issue #256, opened 2018-04-12
Fullscreen stage is blurry


paulkaplan:
I think a few similar issues were closed, but this one is specifically that when on the GUI and going into "fullscreen" mode, the rasterization scale of vector sprites does not change so they look blurry. 

This looks like the same issue, but was closed by the submitter when other svg re-rasterization was merged. https://github.com/LLK/scratch-gui/issues/1041

I think we may want to integrate the ratio of the `nativeSize` (really, scratch coordinate size) and the gl.canvas size into the scale we pass to the SVG renderer. Basically the ratio of nativeSize / gl.canvas size acts in a similar way to "devicePixelRatio", and should be treated like that.

Thoughts @cwillisf?

cwillisf:
Yeah - that sounds good to me. It's a little bit tricky since various parts of the system (render, GUI, VM) need to know relative ratios for physical display pixels, logical display "pixels" (on high DPI displays), Scratch coordinates, and mouse coordinates, but as long as we can do all the necessary conversions I don't have a problem with combining two scaling factors into one variable like you're suggesting.

PR #268, opened 2018-04-30
Scale the drawable scale by the ratio of pixel size to native size.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-render/issues/256
Fixes https://github.com/LLK/scratch-gui/issues/1374

### Proposed Changes

When requesting a drawable texture, combine the drawable scale with the pixel / logical size ratio (large in fullscreen mode, for example). 

### Reason for Changes

This makes the vector sprites not blurry in fullscreen mode

image

image


hedgehog125:
What about pen? It would be nice if that was smooth too. I understand it's stored as bitmap but maybe pen that's drawn while it's in fullscreen could have the full resolution?

PR #272, opened 2018-05-07
Use src/index for browser entry point

mzgoddard:
### Proposed Changes

- Use src/index for browser entry point instead of webpack build library
- Depend on dependencies used in `src` as non-dev dependencies
- Build node version with non-dev dependencies as external modules

### Reason for Changes

Sibling of https://github.com/LLK/scratch-vm/pull/1106

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -21,10 +21,6 @@ const base = {
                 options: {
                     presets: [['env', {targets: {browsers: ['last 3 versions', 'Safari >= 8', 'iOS >= 8']}}]]
                 }
-            },
-            {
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -84,6 +80,14 @@ module.exports = [
             libraryTarget: 'commonjs2',
             path: path.resolve('dist', 'node'),
             filename: '[name].js'
+        },
+        externals: {
+            '!ify-loader!grapheme-breaker': 'grapheme-breaker',
+            '!ify-loader!linebreak': 'linebreak',
```
cwillisf: Ah! I was wondering how to take care of that. TIL!

PR #286, opened 2018-05-16
Remove speech bubbles' reliance on svg quirks mode's _transformText

fsih:
### Resolves
Progress toward fixing https://github.com/LLK/scratch-paint/issues/436

### Proposed Changes
Quirks mode (_transformText and _transformMeasurements in scratch-svg-renderer) deals with fixing really weird stuff scratch 2 did to exported SVGs and should only be applied when converting from scratch 2 to scratch 3 projects. In scratch 3 we want to trust how SVGs are formatted.

Speech bubbles currently rely on quirks mode, so a first step is to remove speech bubbles' dependency on quirks mode. This includes having them be responsible for their own style, instead of expecting quirks mode to enforce a style, and exporting text with tspans instead of \n linebreaks.

### DIFF
--- a/src/util/svg-text-bubble.js
+++ b/src/util/svg-text-bubble.js
@@ -3,6 +3,7 @@ const SvgRenderer = require('scratch-svg-renderer').SVGRenderer;
 
 const MAX_LINE_LENGTH = 170;
 const MIN_WIDTH = 50;
+const STROKE_WIDTH = 4;
 
 class SVGTextBubble {
     constructor () {
@@ -57,7 +58,7 @@ class SVGTextBubble {
                 <path
                   d="${pathString}"
                   stroke="rgba(0, 0, 0, 0.15)"
-                  stroke-width="4"
+                  stroke-width="${STROKE_WIDTH}"
                   fill="rgba(0, 0, 0, 0.15)"
                   stroke-line-join="round"
               />
@@ -113,7 +114,7 @@ class SVGTextBubble {
                     rx="${rx}" ry="${ry}"
                     fill="rgba(0, 0, 0, 0.15)"
                     stroke="rgba(0, 0, 0, 0.15)"
-                    stroke-width="4"
+                    stroke-width="${STROKE_WIDTH}"
                 />
                 <ellipse
                     cx="${cx}" cy="${cy}"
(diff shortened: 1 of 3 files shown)


src/util/svg-text-bubble.js
```diff
@@ -1,17 +1,29 @@
...
+    /**
+     * @return {SVGElement} an SVG text node with the properties that we want for speech bubbles.
+     */
+    makeSvgTextElement () {
+        const svgText = document.createElementNS('http://www.w3.org/2000/svg', 'text');
+        svgText.setAttribute('alignment-baseline', 'text-before-edge');
+        svgText.setAttribute('font-size', '14');
+        svgText.setAttribute('fill', '#575E75');
+        // TODO Do we want to use the new default sans font instead of Helvetica?
+        svgText.setAttribute('font-family', 'Helvetica');
```
paulkaplan: You still need to use `Helvetica, Arial, sans-serif` instead of just Helvetica to subvert the font converter. This PR as written changes the text bubble font to the Helvetica replacement.

src/util/svg-text-bubble.js
```diff
@@ -133,30 +145,28 @@ class SVGTextBubble {
...
-        const svgString = this._wrapSvgFragment(this._textFragment());
+        const svgString = this._wrapSvgFragment(this._textFragment);
         if (!this._textSizeCache[svgString]) {
             this._textSizeCache[svgString] = this.svgRenderer.measure(svgString);
         }
         return this._textSizeCache[svgString];
     }
 
     _wrapSvgFragment (fragment) {
+        // @todo generate view box
```
paulkaplan: Why?

src/util/svg-text-bubble.js
fsih: When https://github.com/LLK/scratch-svg-renderer/pull/22 goes in, hopefully soon after this PR, quirks mode will no longer be run on speech bubbles so it should be safe to switch this back to helvetica.

src/util/svg-text-bubble.js
fsih: Throwing away any existing viewbox and generating a new one is part of quirks mode to fix svgs from scratch 2.0. If it's not too hard, speech bubbles should generate their own view boxes, instead of depending on quirks mode (right now I have to make an exception in svgRenderer so that the view box code still gets run on speech bubbles)

fsih: I added the viewbox to the svg

PR #296, opened 2018-05-29
Remove duplicate xml escaping

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/2148

### Proposed Changes

We used to write the user string directly into the svg string, but now we are setting the `.textContent` of an actual dom node, which does the escaping for us.

### Test Coverage

I added an additional test to the integration tester that does a snapshot test on the svg strings from the bubble creator. This is just to catch any unintended changes to that code, like this one. It had to be in the integration test because it requires the real svg dom to run. Luckily tap has a built in snapshot tester, which is useful. Sadly couldn't change the snapshot path to nest it under the `test` dir the way I'd like :(

But on the plus side yay tests!

BTW I ran the snapshot test without the fix, and the output was pretty readable, showed a diff with the snapshot the revealed the first character, a `<`, was supposed to be `&lt;`, but was actually becomping `&amp;lt;`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 dist/*
 node_modules/*
 playground/*
+tap-snapshots/*
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "uglifyjs-webpack-plugin": "^1.2.5",
     "webpack": "^4.8.0",
     "webpack-cli": "^2.0.15",
-    "webpack-dev-server": "^3.1.4",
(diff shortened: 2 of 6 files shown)


paulkaplan:
Sorry, looks like the font sizing was causing the dimensions to be slightly off on the CI. I changed the test so that it is only looking at the text fragment, to avoid any issues with different fonts being available changing the exact sizing of the bubble. 

@fsih ready for re-review

PR #312, opened 2018-07-13
Touching color implementation

gnarf:
Updated July 19

* Implement CPU side rendering for color effects / alpha blends
* Convert isTouchingColor to use CPU side
* Threshold of a number of pixels that is just too much for the cpu render
* Hybrid approach asks the GPU to do a draw then starts testing using the CPU, when reaching
it's threshold will ask the GPU for the pixels, hopefully spinning both pipelines for even the "worst case"

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -6,7 +6,40 @@ const ShaderManager = require('./ShaderManager');
 const Skin = require('./Skin');
 const EffectTransform = require('./EffectTransform');
 
+/**
+ * An internal workspace for calculating texture locations from world vectors
+ * this is REUSED for memory conservation reasons
+ * @type {twgl.v3}
+ */
 const __isTouchingPosition = twgl.v3.create();
+
+/**
+ * Convert a scratch space location into a texture space float.  Uses the
+ * internal __isTouchingPosition as a return value, so this should be copied
+ * if you ever need to get two local positions and store both.  Requires that
+ * the drawable inverseMatrix is up to date.
+ *
+ * @param {Drawable} drawable The drawable to get the inverse matrix and uniforms from
+ * @param {twgl.v3} vec [x,y] scratch space vector
+ * @return {twgl.v3} [x,y] texture space float vector - transformed by effects and matrix
+ */
+const getLocalPosition = (drawable, vec) => {
+    // Transfrom from world coordinates to Drawable coordinates.
+    const localPosition = __isTouchingPosition;
+    const v0 = vec[0];
+    const v1 = vec[1];
+    const m = drawable._inverseMatrix;
+    // var v2 = v[2];
+    const d = (v0 * m[3]) + (v1 * m[7]) + m[15];
+    // The RenderWebGL quad flips the texture's X axis. So rendered bottom
+    // left is 1, 0 and the top right is 0, 1. Flip the X axis so
+    // localPosition matches that transformation.
+    localPosition[0] = 0.5 - (((v0 * m[0]) + (v1 * m[4]) + m[12]) / d);
+    localPosition[1] = (((v0 * m[1]) + (v1 * m[5]) + m[13]) / d) + 0.5;
+    // Apply texture effect transform.
+    EffectTransform.transformPoint(drawable, localPosition, localPosition);
+    return localPosition;
+};
 
 class Drawable {
     /**
@@ -381,36 +414,7 @@ class Drawable {
             return false;
         }
 
-        if (this._transformDirty) {
-            this._calculateTransform();
-        }
-
-        // Get the inverse of the model matrix or update it.
-        const inverse = this._inverseMatrix;
-        if (this._inverseTransformDirty) {
-            const model = twgl.m4.copy(this._uniforms.u_modelMatrix, inverse);
-            // The normal matrix uses a z scaling of 0 causing model[10] to be
-            // 0. Getting a 4x4 inverse is impossible without a scaling in x, y,
-            // and z.
-            model[10] = 1;
-            twgl.m4.inverse(model, model);
-            this._inverseTransformDirty = false;
-        }
-
-        // Transfrom from world coordinates to Drawable coordinates.
-        const localPosition = twgl.m4.transformPoint(inverse, vec, __isTouchingPosition);
-
-        // Transform into texture coordinates. 0, 0 is the bottom left. 1, 1 is
-        // the top right.
-        localPosition[0] += 0.5;
-        localPosition[1] += 0.5;
-        // The RenderWebGL quad flips the texture's X axis. So rendered bottom
-        // left is 1, 0 and the top right is 0, 1. Flip the X axis so
-        // localPosition matches that transformation.
-        localPosition[0] = 1 - localPosition[0];
-
-        // Apply texture effect transform.
-        EffectTransform.transformPoint(this, localPosition, localPosition);
+        const localPosition = getLocalPosition(this, vec);
 
         if (this.useNearest) {
             return this.skin.isTouchingNearest(localPosition);
@@ -460,6 +464,7 @@ class Drawable {
         bounds.initFromPointsAABB(transformedHullPoints);
         return bounds;
     }
+
     /**
      * Get the precise bounds for the upper 8px slice of the Drawable.
      * Used for calculating where to position a text bubble.
@@ -514,6 +519,7 @@ class Drawable {
      * @return {!Rectangle} Bounds for the Drawable.
      */
     getFastBounds () {
+        this.updateMatrix();
         if (!this.needsConvexHullPoints()) {
             return this.getBounds();
         }
@@ -543,6 +549,27 @@ class Drawable {
             transformedHullPoints.push(glPoint);
         }
         return transformedHullPoints;
+    }
+
+    /**
+     * Update the transform matrix and calculate it's inverse for collision
+     * and local texture position purposes.
+     */
+    updateMatrix () {
+        if (this._transformDirty) {
+            this._calculateTransform();
+        }
+        // Get the inverse of the model matrix or update it.
+        if (this._inverseTransformDirty) {
+            const inverse = this._inverseMatrix;
+            twgl.m4.copy(this._uniforms.u_modelMatrix, inverse);
+            // The normal matrix uses a z scaling of 0 causing model[10] to be
+            // 0. Getting a 4x4 inverse is impossible without a scaling in x, y,
+            // and z.
+            inverse[10] = 1;
+            twgl.m4.inverse(inverse, inverse);
+            this._inverseTransformDirty = false;
+        }
     }
 
     /**
@@ -585,6 +612,28 @@ class Drawable {
         id |= (g & 255) << 8;
         id |= (b & 255) << 16;
         return id + RenderConstants.ID_NONE;
+    }
+
+    /**
+     * Sample a color from a drawable's texture.
+     * @param {twgl.v3} vec The scratch space [x,y] vector
+     * @param {Drawable} drawable The drawable to sample the texture from
+     * @param {Uint8ClampedArray} dst The "color4b" representation of the texture at point.
+     * @returns {Uint8ClampedArray} The dst object filled with the color4b
+     */
+    static sampleColor4b (vec, drawable, dst) {
+        const localPosition = getLocalPosition(drawable, vec);
+        if (localPosition[0] < 0 || localPosition[1] < 0 ||
+            localPosition[0] > 1 || localPosition[1] > 1) {
+            dst[3] = 0;
+            return dst;
+        }
+        const textColor =
+        // commenting out to only use nearest for now
+        // drawable.useNearest ?
+             drawable.skin._silhouette.colorAtNearest(localPosition, dst);
+        // : drawable.skin._silhouette.colorAtLinear(localPosition, dst);
+        return EffectTransform.transformColor(drawable, textColor, textColor);
     }
 }
 
--- a/src/EffectTransform.js
+++ b/src/EffectTransform.js
@@ -20,7 +20,165 @@ const CENTER_X = 0.5;
  */
 const CENTER_Y = 0.5;
 
+// color conversions grabbed from https://gist.github.com/mjackson/5311256
+
+/**
+ * Converts an RGB color value to HSL. Conversion formula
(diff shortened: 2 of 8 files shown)


src/RenderWebGL.js
```diff
@@ -656,91 +661,30 @@ class RenderWebGL extends EventEmitter {
...
+        const point = __isTouchingDrawablesPoint;
+        for (let x = bounds.left; x <= bounds.right; x++) {
+            // Scratch Space - +y is top
+            point[0] = x;
+            for (let y = bounds.bottom; y <= bounds.top; y++) {
+                point[1] = y;
+                // our point matches parent point
+                if (drawable.isTouching(point) && (!mask3b || maskMatches(drawable.sampleColor4b(point), mask3b))) {
+                    const topTarget = candidates[0].drawable;
+                    const rest = candidates.slice(1).map(({drawable: d}) => d);
```
gnarf: can take these two consts out of the tight loop, might help

src/Silhouette.js
```diff
@@ -27,6 +27,17 @@ const getPoint = ({_width: width, _height: height, _data: data}, x, y) => {
     return data[(y * width) + x];
 };
 
+const _default4b = new Uint8ClampedArray(4);
+
+const getColor4b = ({_width: width, _height: height, _colorData: data}, x, y) => {
```
mzgoddard: Pass a `dst` argument that defaults to a `new Uint8ClampedArray(4)` and set its indices manually. The default I think could be replaced with `.fill(0)`.

src/Silhouette.js
```diff
@@ -67,12 +79,54 @@ class Silhouette {
...
 
         this._data = new Uint8ClampedArray(imageData.data.length / 4);
+        this._colorData = imageData.data;
 
         for (let i = 0; i < imageData.data.length; i += 4) {
             this._data[i / 4] = imageData.data[i + 3];
         }
     }
 
+    colorAtNearest (vec) {
```
mzgoddard: Pass a `dst` option without a default to `getColor4b`.

src/Silhouette.js
```diff
@@ -67,12 +79,54 @@ class Silhouette {
...
+        }
+
+        return getColor4b(
+            this,
+            Math.floor(vec[0] * (this._width - 1)),
+            Math.floor(vec[1] * (this._height - 1))
+        );
+    }
+
+    colorAtLinear (vec) {
```
mzgoddard: Take a `dst` option that defaults to a `new Uint8ClampedArray(4)`.

src/Silhouette.js
```diff
@@ -67,12 +79,54 @@ class Silhouette {
...
+        const yFloor = Math.floor(y);
+
+        const corners = [
+            getColor4b(this, xFloor, yFloor),
+            getColor4b(this, xFloor + 1, yFloor),
+            getColor4b(this, xFloor, yFloor + 1),
+            getColor4b(this, xFloor + 1, yFloor + 1)
+        ];
+
+        return Uint8ClampedArray.from({length: 4}, (_, i) => (
```
mzgoddard: Most VMs I believe will allocate a closure object here. Manually setting the indices with a helper function will be faster.

src/Silhouette.js
```diff
@@ -67,12 +79,54 @@ class Silhouette {
...
+        const x = vec[0] * (this._width - 1);
+        const y = vec[1] * (this._height - 1);
+
+        const xDelta = x % 1;
+        const yDelta = y % 1;
+
+        const xFloor = Math.floor(x);
+        const yFloor = Math.floor(y);
+
+        const corners = [
```
mzgoddard: Assign values to a static array instance.

src/Silhouette.js
```diff
@@ -67,12 +88,57 @@ class Silhouette {
...
+        this._colorData = imageData.data;
 
         for (let i = 0; i < imageData.data.length; i += 4) {
             this._data[i / 4] = imageData.data[i + 3];
         }
     }
 
+    colorAtNearest (vec, dst) {
+        dst = dst || new Uint8ClampedArray(4);
+        if (!this._colorData) {
```
mzgoddard: I wonder if it would be faster to define two versions of this function. One that returns `dst.fill` and one that returns `getColor4b`. Set `colorAtNearest` to the right implementation when initializing `Silhouette` and updating it.

src/Silhouette.js
```diff
@@ -27,6 +27,26 @@ const getPoint = ({_width: width, _height: height, _data: data}, x, y) => {
...
 };
 
+const __cornerWork = [
+    new Uint8ClampedArray(4),
+    new Uint8ClampedArray(4),
+    new Uint8ClampedArray(4),
+    new Uint8ClampedArray(4)
+];
+
+const getColor4b = ({_width: width, _height: height, _colorData: data}, x, y, dst = new Uint8ClampedArray(4)) => {
```
mzgoddard: `dst = dst || new Uint8ClampedArray(4)`

src/RenderWebGL.js
```diff
@@ -1495,6 +1518,51 @@ class RenderWebGL extends EventEmitter {
         // Simplify boundary points using convex hull.
         return hull(boundaryPoints, Infinity);
     }
+
+    static sampleColor4b (vec, drawable, dst) {
+        const localPosition = Drawable.getLocalPosition(drawable, vec);
```
mzgoddard: I wonder if there would be an improvement by listing a separate copy of `getLocalPosition` in this file. I think this may mean this method would be better suited on `Drawable` than `RenderWebGL`.

src/RenderWebGL.js
```diff
@@ -653,94 +649,118 @@ class RenderWebGL extends EventEmitter {
...
+            point[0] = x;
+            for (let y = bounds.bottom; y <= bounds.top; y++) {
+                point[1] = y;
+                if (
+                    // if we use a mask, check our sample color
+                    (hasMask ?
+                        maskMatches(RenderWebGL.sampleColor4b(point, drawable, color), mask3b) :
+                        drawable.isTouching(point)) &&
+                    // and the target color is drawn at this pixel
+                    colorMatches(RenderWebGL.sampleColor3b(point, candidates, color), color3b, 0)
```
mzgoddard: Save a reference to this function in the `const` group before the `for` loop?

mzgoddard:
Here are some numbers from my test devices:

|  Test | Platform | Before (blocks/s) | After (blocks/s) | Change (percent) |
|  ------ | ------ | -----: | -----: | -----: |
|  219313833 | MBP 15" 2017 Chrome | 103050 | 131017 | 27.13% |
|   | MBP 15" 2017 Safari | 87191 | 136301 | 56.32% |
|   | MBP 15" 2017 Firefox | 93468 | 129096 | 38.11% |
|   | SMS Chromebook 2017 | 31439 | 38936 | 23.84% |
|   | iPad Mini 2015 | 20482 | 29590 | 44.46% |
|  236115215 | MBP 15" 2017 Chrome | 6119 | 12967 | 111.91% |
|   | MBP 15" 2017 Safari | 9575 | 16710 | 74.51% |
|   | MBP 15" 2017 Firefox | 3515 | 6856 | 95.04% |
|   | SMS Chromebook 2017 | 2237 | 4105 | 83.50% |
|   | iPad Mini 2015 | 3158 | 4602 | 45.72% |

src/RenderWebGL.js
```diff
@@ -32,26 +34,40 @@ const __candidatesBounds = new Rectangle();
...
+    // has some non-alpha component to test against
+    a[3] > 0 &&
+    (a[0] & 0b11111100) === (b[0] & 0b11111100) &&
+    (a[1] & 0b11111100) === (b[1] & 0b11111100) &&
+    (a[2] & 0b11111100) === (b[2] & 0b11111100)
+);
+
+/**
+ * Determines if the given color is "close enough" (only test the 5 top bits for
+ * red and green, 5 bits for blue).  These bit masks are what scratch 2 used to use,
```
cwillisf: typo: should be `4 bits for blue`

src/RenderWebGL.js
```diff
@@ -653,94 +669,115 @@ class RenderWebGL extends EventEmitter {
...
             }
+
+            return false;
         }
 
+        const drawable = this._allDrawables[drawableID];
+        const point = __isTouchingDrawablesPoint;
+        const color = __touchingColor;
+        const hasMask = Boolean(mask3b);
+        for (let x = bounds.left; x <= bounds.right; x++) {
```
cwillisf: There's a chance that this would be more cache-friendly by swapping the loops so that `x` is the inner loop. I suspect there won't be a measurable difference, especially with a size cap on the CPU-rendered path, but you never know...

src/Silhouette.js
```diff
@@ -67,12 +101,65 @@ class Silhouette {
         const imageData = ctx.getImageData(0, 0, width, height);
 
         this._data = new Uint8ClampedArray(imageData.data.length / 4);
+        this._colorData = imageData.data;
+        // delete our custom overriden "uninitalized" color functions
+        // let the prototype work for itself
+        delete this.colorAtNearest;
+        delete this.colorAtLinear;
```
cwillisf: I hadn't seen this technique before... kinda fun :)

src/RenderWebGL.js
```diff
@@ -16,6 +16,13 @@ const log = require('./util/log');
 
 const __isTouchingDrawablesPoint = twgl.v3.create();
 const __candidatesBounds = new Rectangle();
+const __touchingColor = new Uint8ClampedArray(4);
+const __blendColor = new Uint8ClampedArray(4);
+
+// More pixels than this and we give up to the GPU and take the cost of readPixels
+// Width * Height * Number of drawables at location
+const __cpuTouchingColorPixelCount = location.href.includes('count=') ?
+    Number(location.href.match(/count=([^&#]+)/)[1]) : 1000000;
```
gnarf: TODO: Remove this query param - but this should let us play with benchmarks / thresholds:

I.E: http://localhost:8073/playground/index.html?count=1e4#236115215 

anything more than 1000 pixels (width * height * (candidates.length + 1 (ourself) )) would get thrown to GPU

src/RenderWebGL.js
```diff
@@ -716,9 +768,12 @@ class RenderWebGL extends EventEmitter {
             gl.colorMask(true, true, true, true);
             gl.disable(gl.STENCIL_TEST);
         }
+    }
 
-        const pixels = new Uint8Array(Math.floor(bounds.width * bounds.height * 4));
-        gl.readPixels(0, 0, bounds.width, bounds.height, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    _isTouchingColorGpuFin (bounds, color3b) {
+        const gl = this._gl;
+        const pixels = new Uint8Array(Math.floor(bounds.width * (bounds.height - stop) * 4));
```
mzgoddard: Where does `stop` come from?

gnarf:
Just rebased

gnarf:
Interesting --- the new pick integration tests I wrote uncovered a bug with pick in the rewrite! How nice!

src/RenderWebGL.js
```diff
@@ -1556,6 +1620,43 @@ class RenderWebGL extends EventEmitter {
...
+    /**
+     * Sample a "final" color from an array of drawables at a given scratch space.
+     * Will blend any alpha values with the drawables "below" it.
+     * @param {twgl.v3} vec Scratch Vector Space to sample
+     * @param {Array<Drawables>} drawables A list of drawables with the "top most"
+     *              drawable at index 0
+     * @param {Uint8ClampedArray} dst The color3b space to store the answer in.
+     * @return {Uint8ClampedArray} The dst vector with everything blended down.
+     */
+    static sampleColor3b (vec, drawables, dst) {
```
cwillisf: It looks like this function is relying on the caller to filter out hidden drawables, if desired. Is that accurate / intended?

src/Drawable.js
```diff
@@ -585,6 +612,28 @@ class Drawable {
...
+        if (localPosition[0] < 0 || localPosition[1] < 0 ||
+            localPosition[0] > 1 || localPosition[1] > 1) {
+            dst[3] = 0;
+            return dst;
+        }
+        const textColor =
+        // commenting out to only use nearest for now
+        // drawable.useNearest ?
+             drawable.skin._silhouette.colorAtNearest(localPosition, dst);
+        // : drawable.skin._silhouette.colorAtLinear(localPosition, dst);
```
cwillisf: Could you file an issue to track why `colorAtLinear` is disabled / what we need to do if we want to enable it / why we should or shouldn't do so?

test/integration/index.html
```diff
@@ -11,6 +11,7 @@
     <script>
         // These variables are going to be available in the "window global" intentionally.
         // Allows you easy access to debug with `vm.greenFlag()` etc.
+        window.devicePixelRatio = 1;
```
cwillisf: Why?

src/EffectTransform.js
```diff
@@ -56,7 +214,7 @@ class EffectTransform {
             const offsetX = dst[0] - CENTER_X;
             const offsetY = dst[1] - CENTER_Y;
             // float offsetMagnitude = length(offset);
-            const offsetMagnitude = twgl.v3.length(dst);
+            const offsetMagnitude = Math.sqrt(Math.pow(offsetX, 2) + Math.pow(offsetY, 2));
```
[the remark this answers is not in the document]
gnarf: nice thats useful - i'll check into it - i decided not to use `offset ** 2` either ;)

src/RenderWebGL.js
gnarf: Yes and no - this is called inside of a tight loop, so I got rid of the visibility checks. 
The redundant checks are just wasted compute this deep.  I should probably document that it will work fine with any value of ghost, including 100, but not hidden/visible -- if "hide" it would still render here

test/integration/index.html
gnarf: just to force a specific value, we could lock it to 2 if we wanted instead?

src/Drawable.js
gnarf: so it's like 5* the math to sample linear (and i think the math might actually be buggy for alpha pixels) - that's mostly why

test/integration/index.html
cwillisf: 1 is probably more common right now, so it probably makes sense for our tests. Does setting this actually impact the way the canvas is rendered?

PR #314, opened 2018-07-16
offset pen lines of width 1 and 3

mzgoddard:
### Resolves

https://github.com/LLK/scratch-vm/issues/1306

### Proposed Changes

Offset pen lines with width of 1 or 3.

### Reason for Changes

Lines with width 1 or 3 in Scratch 2 seem to be rendered in the center of their pixels leaving crisp edges for vertical and horizontal lines. Lines with odd widths larger than 3 have aliased lines like the lines in Scratch 3 already are.

### Test Coverage

The images below are differences of Scratch 2 and Scratch 3 rendering https://llk.github.io/scratch-gui/develop/#236563663. The second image shows the difference with this change.

screen shot 2018-07-16 at 3 22 34 pm

screen shot 2018-07-17 at 1 09 37 pm

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -132,9 +132,13 @@ class PenSkin extends Skin {
     drawLine (penAttributes, x0, y0, x1, y1) {
         const ctx = this._canvas.getContext('2d');
         this._setAttributes(ctx, penAttributes);
+
+        // Width 1 and 3 lines need to be offset by 0.5.
+        const diameter = penAttributes.diameter || DefaultPenAttributes.diameter;
+        const offset = (Math.max(4 - diameter, 0) % 2) / 2;
         ctx.beginPath();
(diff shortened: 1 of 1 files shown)


cwillisf:
That's a great improvement! Out of curiosity, what other offset schemes did you try? For example, I imagine you tried always offsetting by 0.5, but did you try 0.25 or 0.375? I'm particularly interested in 0.375 since that offset tends to fix quite a few pixel edge/center conflicts in my experience.
All that said, if this is the simplest solution that makes this lines look right then it works for me :)

mzgoddard:
@cwillisf here are some differences with other offsets:

0.125

screen shot 2018-07-23 at 1 28 23 pm

0.25

screen shot 2018-07-23 at 1 30 01 pm

0.375

screen shot 2018-07-23 at 1 29 44 pm

0.625

screen shot 2018-07-23 at 1 28 53 pm

0.5 seems to be the best among this resolution of offsets.

PR #325, opened 2018-08-01
Optimizing isTouching while creating a drawableTouches for sensing mouse pointer

### LINKED ISSUE #284: "Touching mouse pointer" always returns false when the sprite is hidden
### Expected Behavior

In Scratch 2.0 it returns true if it's over the sprite, even if the sprite is hidden.

To test, add and run these blocks:
image

Then hide the sprite, mouseover it, and show it again.  It should have moved.

---
Moved from LLK/scratch-vm#1003
/cc @rachel-fenichel

gnarf:
### Resolves

Resolves #284 (builds on top of @kchadha #322)

### Proposed Changes

Drawable#isTouching no longer has a visibility check, instead `isTouchingDrawables` and other locations that use that method do their own pre-checks to still meet the rules of "touching" but letting pick (mouse-pointer) find an invisible sprite (and um... wow?)

### Reason for Changes

Compatibility with 2.0.

### Test Coverage

Manual testing with #223437838


src/RenderWebGL.js
```diff
@@ -757,13 +761,15 @@ class RenderWebGL extends EventEmitter {
...
+     * @param {?Array<int>} candidateIDs The Drawable IDs to check, otherwise all visible drawables in the renderer
      * @returns {boolean} True if the Drawable is touching one of candidateIDs.
      */
     isTouchingDrawables (drawableID, candidateIDs = this._drawList) {
-
-        const candidates = this._candidatesTouching(drawableID, candidateIDs);
-        if (candidates.length === 0) {
+        const candidates = this._candidatesTouching(drawableID,
+            // even if passed an invisible drawable, we will NEVER touch it!
+            candidateIDs.filter(id => this._allDrawables[id]._visible));
```
kchadha: Should this actually just be _visibleDrawList? I'm actually not sure if I understand the old comment for this function in the context of the old code. Was candidateIDs supposed to be an optional param but then it got turned into a required param? If it is a required param, then neither of these sentences makes sense to me:

> -The Drawable IDs to check, otherwise all drawables in the renderer

> +The Drawable IDs to check, otherwise all visible drawables in the renderer

src/RenderWebGL.js
```diff
@@ -794,26 +800,22 @@ class RenderWebGL extends EventEmitter {
...
+    pick (centerX, centerY, touchWidth, touchHeight, candidateIDs = this._visibleDrawList) {
         const gl = this._gl;
 
         touchWidth = touchWidth || 1;
         touchHeight = touchHeight || 1;
-        candidateIDs = (candidateIDs || this._drawList).filter(id => {
-            const drawable = this._allDrawables[id];
-            const uniforms = drawable.getUniforms();
-            return drawable.getVisible() && uniforms.u_ghost !== 0;
-        });
```
kchadha: Looks like you can now pick up ghosted sprites, I think this could cause compatibility issues with projects..

src/RenderWebGL.js
gnarf: So the problem here is you can pass it an invisible drawable, `isTouchingDrawables` should never touch an invisible drawable - Even if you say "am i touching this hidden drawable" the answer is "no" -- The unit tests actually test for this case, so I'm confident it's doing the right thing

src/RenderWebGL.js
gnarf: Yeah - This is definitely a change, but it's a change that I'm wondering if we shouldn't do.  Basically every other collision detect DOES NOT ignore 100% ghosted sprites - What actually uses `pick` (click on stage? etc) I would love to get some "units" written that test these edges and make sure we have the right behavior before moving on with it.  I'd be happy to update this again to default ignore ghosted sprites.

gnarf: I just reverted the behavior here.

cwillisf:
Hoping to resolve some confusion:

I think the problem arose because the `pick` function, originally designed for mouse picking with the "when sprite clicked" block and for dragging sprites around the stage, started serving more than just that purpose. That and the fact that `_visible` is a sometimes-confusing name for the flag controlled by the "hide" and "show" blocks :)

In general we should match Scratch 2, which is not necessarily what one might expect in all cases. I just tested to be super-double-sure, and below is what I believe is the expected behavior.

These tests were done with this configuration of sprites, called `Sprite1` and `Apple`; note that `Sprite1` is overlapping the stage edge.
image

## Blocks affecting the query sprite

*If the block along the top runs on `Sprite1`, is it possible for the query or action in the left column to succeed?*

| | image | image
---- | ---- | ----
Drag sprite with mouse | no | no
image | no | no
image | yes | yes
image | yes | yes
image | no | yes
image | yes | yes
image | yes | yes

## Blocks affecting another sprite

*If the block along the top runs on `Apple` and the query in the left column runs on `Sprite1`, is it still possible for the query to succeed?*

| | image | image
---- | ---- | ----
image | no | yes
image | no | no
image | no | no

Issue #344, opened 2018-09-13
Infinite loop in WebGLRenderer


mraerino:
### Expected Behavior

When loading a sprite, it is displayed and can be used.

### Actual Behavior

When loading a sprite, the renderer will sometime get into a infinite loop and cause the tab to become unresponsive. The sprite is already displayed at that time.

### Steps to Reproduce

I was loading [this svg](https://gist.github.com/mraerino/ec23aa4ae759f4719f4cc0342932faad) into scratch-paint when the browser became unresponsive. This also happend with JPGs and other SVG sprites before.
When pausing Javascript, the browser was stuck at `RenderWebGL.js:962`. The `bounds` constant in that function had these values:
```
{
  bottom: -Infinity
  left: Infinity
  right: Infinity
  top: -Infinity
}
```

### Operating System and Browser

MacOS 10.13.6 / Chrome 69

---

**Note:**
I will try to find the cause of the bug myself, since we are basing a project on scratch-gui and need a fix soon, but already wanted to let you know about it. So you can probably expect a pull-request for this.

thisandagain:
Resolved in #349

PR #349, opened 2018-09-24
Fix bug when bounds would be infinite and cause never ending loop

mraerino:
### Resolves

https://github.com/LLK/scratch-render/issues/344

### Proposed Changes

This will add a check when picking an object at a certain point to return early when the bounds are unset for any reason.

### Reason for Changes

Sometimes the `scratch-gui` tab would hang completely when using the `scratch-paint` vector editor.
I tracked this back to the `pick` method in this project.

This is not really easy to reproduce but seems to fix the issue and should be a reasonable improvement in the code.

### Test Coverage

None. This is really hard to reproduce in a test.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -968,7 +968,12 @@ class RenderWebGL extends EventEmitter {
         if (candidateIDs.length === 0) {
             return false;
         }
+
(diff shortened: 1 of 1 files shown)


thisandagain:
@cwillisf Could you take a look at this?

PR #354, opened 2018-10-04
Fix touching mouse

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-vm/issues/1500

### Proposed Changes

- Modify pick tests to reveal that the drawableTouching tests alone will fail (they pass currently due to positioning of the tests in scratch space (the center of the drawn/tested region) and cached matrices from the earlier pick tests.
- Update drawable matrix and silhouette in drawableTouching

### Reason for Changes

drawableTouching is a collision detection method. It's siblings call updateMatrix updateSilhouette so isTouching can be up to date without needing to make extraneous calls to updateMatrix and updateSilhouette itself.

### Test Coverage

Split up the existing pick tests to test this regression.


thisandagain:
@mzgoddard @ktbee Bump

mzgoddard:
Fixed the lint issues.

PR #376, opened 2018-12-12
Fix extraction of drawable to not clip bounds.

evhan55:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/1808

### Notes

I am new to `scratch-render` so not 100% sure if `extractDrawable` is used somewhere else where the clipping was actually intended.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1038,7 +1038,6 @@ class RenderWebGL extends EventEmitter {
      * @return {?DrawableExtraction} Data about the picked drawable
      */
     extractDrawable (drawableID, x, y) {
-        // console.log('here');
         this._doExitDrawRegion();
 
         const drawable = this._allDrawables[drawableID];
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1065,14 +1065,9 @@ class RenderWebGL extends EventEmitter {
         const bufferInfo = twgl.createFramebufferInfo(gl, attachments, clampedWidth, clampedHeight);
 
         // If the new bufferInfo is invalid, fall back to using the smaller _queryBufferInfo
-        console.log(bufferInfo);
-        if (gl.checkFramebufferStatus(bufferInfo.framebuffer) === 0) {
-            // twgl.bindFramebufferInfo(gl, this._queryBufferInfo);
-            twgl.bindFramebufferInfo(gl, bufferInfo);
-            console.log(bufferInfo);
-        } else {
-            twgl.bindFramebufferInfo(gl, bufferInfo);
-            console.log(bufferInfo);
+        twgl.bindFramebufferInfo(gl, bufferInfo);
+        if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) !== gl.FRAMEBUFFER_COMPLETE) {
+            twgl.bindFramebufferInfo(gl, this._queryBufferInfo);
         }
 
         // Translate to scratch units relative to the drawable
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1038,6 +1038,7 @@ class RenderWebGL extends EventEmitter {
      * @return {?DrawableExtraction} Data about the picked drawable
      */
     extractDrawable (drawableID, x, y) {
+        // console.log('here');
         this._doExitDrawRegion();
 
         const drawable = this._allDrawables[drawableID];
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1052,7 +1052,7 @@ class RenderWebGL extends EventEmitter {
         const bounds = drawable.getFastBounds();
         bounds.snapToInt();
 
-        // Use a new bufferInfo since _queryBufferInfo is limited to 480x360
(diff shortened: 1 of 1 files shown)


fsih:
Slowly backing away because I don't know this code

cwillisf:
CC @rschamp, in case you're interested

src/RenderWebGL.js
```diff
@@ -1040,11 +1040,18 @@ class RenderWebGL extends EventEmitter {
...
 
         const bounds = drawable.getFastBounds();
         bounds.snapToInt();
 
+        // Use a new bufferInfo since this._queryBufferInfo is limited to 480x360
+        const attachments = [
+            {format: gl.RGBA},
+            {format: gl.DEPTH_STENCIL}
+        ];
+        const bufferInfo = twgl.createFramebufferInfo(gl, attachments, bounds.width, bounds.height);
```
cwillisf: This could fail if the bounds are too big, and even if it succeeds it could eat up a lot of memory. I would suggest checking the max texture size with something like `const maxTextureSize = gl.getParameter(gl.MAX_TEXTURE_SIZE);` and then using something like `Math.min(2048, bounds.width, maxTextureSize)` and similar for `bounds.height`. The 2048 is in there to provide a reasonable maximum even if you have a huge sprite and an ambitious GPU :)

cwillisf: I would also add a `gl.checkFramebufferStatus` check on the new `bufferInfo` and, if that fails, fall back to using the query buffer as a last resort.

src/RenderWebGL.js
```diff
@@ -1040,11 +1040,30 @@ class RenderWebGL extends EventEmitter {
...
+        // Make a new bufferInfo since this._queryBufferInfo is limited to 480x360
+        const attachments = [
+            {format: gl.RGBA},
+            {format: gl.DEPTH_STENCIL}
+        ];
+        const bufferInfo = twgl.createFramebufferInfo(gl, attachments, clampedWidth, clampedHeight);
+
+        // If the new bufferInfo is invalid fall back to using the smaller _queryBufferInfo
+        if (gl.checkFramebufferStatus(bufferInfo.framebuffer) === 0) {
+            // twgl.bindFramebufferInfo(gl, this._queryBufferInfo);
```
cwillisf:
```suggestion
            twgl.bindFramebufferInfo(gl, this._queryBufferInfo);
```

src/RenderWebGL.js
```diff
@@ -1040,11 +1040,30 @@ class RenderWebGL extends EventEmitter {
...
+            twgl.bindFramebufferInfo(gl, bufferInfo);
```
cwillisf:
```suggestion
```

src/RenderWebGL.js
evhan55: Oh yes!  This is why I hadn't re-assigned you yet.  I was testing this code, but the check failed.  So then I went to read up some WebGL basics to try to understand why, and I still haven't figured it out.  Let me know if you have any ideas!  I pushed the commits as a habit to not lose work in progress.

evhan55: `if (gl.checkFramebufferStatus(bufferInfo.framebuffer) === 0)` does not pass

evhan55: Hmm, it might have to do with the order in which I am binding/checking the frame buffers, am investigating!

evhan55:
@cwillisf: I fixed the `checkFramebufferstatus` check to happen after binding, and to check the right target for (what I think is) the correct status value.

I don't know how to test the "non-success" path, I tried on a super large sprite: `#251658287`, but the new, larger bufferInfo still works on my graphics card and I get a huge extracted drawable for it.

dec-19-2018 12-06-16

**Basic Test:**
1. New project
2. Set default sprite size to 600%
3. Click on sprite to drag out of stage

**Result:**
Should see entire sprite

cwillisf:
I think the easiest way to test the query buffer fallback is to temporarily modify the code to cause the new buffer to fail. Maybe something like this:
```js
const bufferInfo = twgl.createFramebufferInfo(gl, attachments, 99999*clampedWidth, clampedHeight);
```

evhan55:
@cwillisf Oh cool that did it!  Resulting clipped buffer + WebGL warnings to the right:

screen shot 2018-12-19 at 1 03 29 pm

evhan55:
@cwillisf Thanks for all the guidance, I learned some new things!

Sounds good, that means I shouldn't merge until after the release?

evhan55:
@cwillisf how does a PR like this go about getting unblocked?

src/RenderWebGL.js
```diff
@@ -1030,6 +1030,7 @@ class RenderWebGL extends EventEmitter {
      * @return {?DrawableExtraction} Data about the picked drawable
      */
     extractDrawable (drawableID, x, y) {
+        // console.log('here');
```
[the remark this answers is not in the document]
evhan55: Yes, will remove!

evhan55:
@cwillisf Ok I removed the debug comment and re-tested this code with the latest scratch-vm and scratch-gui from develop.  Ready for re-review.

PR #399, opened 2019-01-16
Change brightness effect to match Scratch 2.0 in 2D mode

### LINKED ISSUE #142: Brightness + Color effect doesn't work the same as Scratch 2
### Expected Behavior

_Please describe what should happen_
Scratch 2, if you set brightness effect to 100 and color effect to 50, this is what you get
image

### Actual Behavior

_Describe what actually happens_

In Scratch 3, you get all white (I added a background so that you can see it). 
image

---

This bug is a simplification of a problem I had when opening this project in scratch 3 https://scratch.mit.edu/projects/1270935/
It should look like this
image

And instead looked like this 
image

---
cc @cwillisf

cwillisf:
I wasn't sure who would want to review a GLSL change so I just assigned a bunch of people. Feel free to unassign yourself if you're not interested :)

If you're unfamiliar with GLSL or just need a refresher, take a look at pages 3 & 4 of this PDF:
https://www.khronos.org/files/webgl/webgl-reference-card-1_0.pdf

### Resolves

Resolves #142

### Proposed Changes

Previously this repository's brightness effect matched the behavior of Scratch 2.0's 3D renderer by adjusting the value component of the HSV representation of a color. Now, it matches the behavior of Scratch 2.0's 2D renderer, which simply offsets (and clamps/saturates) the RGB values.

Note that to fully match Scratch 2.0's 2D renderer the brightness effect must happen after the color effect, not before.

### Reason for Changes

It appears that most uses of the brightness effect in Scratch 2.0 are not paired with other graphical effects which cause the 3D renderer to activate, so matching the 2D renderer will likely cause fewer projects to display incorrectly in Scratch 3.0.

### Test Coverage

Testing is easiest by comparing the Scratch 2.0 renderer against the scratch-render playground. See also #142 for example projects to test.

### DIFF
--- a/src/shaders/sprite.frag
+++ b/src/shaders/sprite.frag
@@ -45,13 +45,13 @@ uniform sampler2D u_skin;
 
 varying vec2 v_texCoord;
 
-#if !defined(DRAW_MODE_silhouette) && (defined(ENABLE_color) || defined(ENABLE_brightness))
+#if !defined(DRAW_MODE_silhouette) && (defined(ENABLE_color))
 // Branchless color conversions based on code from:
 // http://www.chilliant.com/rgb2hsv.html by Ian Taylor
 // Based in part on work by Sam Hocevar and Emil Persson
 // See also: https://en.wikipedia.org/wiki/HSL_and_HSV#Formal_derivation
 
-// Smaller values can cause problems with "color" and "brightness" effects on some mobile devices
+// Smaller values can cause problems on some mobile devices
 const float epsilon = 1e-3;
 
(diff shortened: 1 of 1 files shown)


cwillisf:
Note that adding `?w=1` to the URL of the diff view will tell GitHub's diff to ignore whitespace differences and make the diff much more readable for this change :)

cwillisf:
After talking with both @paulkaplan and @rschamp I think the right move is to merge this for now and then consider more complicated solutions if we receive a lot of complaint.

Issue #396, opened 2019-01-10
Fence width logic differs between Scratch 2 and Scratch 3


ktbee:
### Expected Behavior

The fence width and height should be the same for Scratch 2 and 3

### Actual Behavior

`getFencedPositionOfDrawable` in RenderWebGL.js sets an inset for [width](https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L1305) and [height](https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L1311) instead of one inset for both like [Scratch 2 uses.](https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchSprite.as#L209)

### Steps to Reproduce

Set a sprite that is taller than it is wide to go off the edge of the stage. You'll observe that it goes further off the stage in Scratch 2 than 3.

Example Scratch 3 project: https://scratch.mit.edu/projects/278625596

![screen shot 2019-01-10 at 1 37 27 pm](https://user-images.githubusercontent.com/7991694/50989280-ee024e80-14dc-11e9-99a7-d64e83acbc9f.png)

Original sb2 for that project

[inset-test.sb2.zip](https://github.com/LLK/scratch-render/files/2746638/inset-test.sb2.zip)

![screen shot 2019-01-10 at 1 38 19 pm](https://user-images.githubusercontent.com/7991694/50989322-05413c00-14dd-11e9-9d1b-060bc0cf6cdb.png)

[This could be related to why you don't fall into pits!](https://github.com/LLK/scratch-vm/issues/1451)

thisandagain:
Very likely the root cause of https://github.com/LLK/scratch-vm/issues/1451

ktbee:
Resolved by https://github.com/LLK/scratch-render/pull/400

PR #400, opened 2019-01-16
Add inset logic that matches Scratch 2

ktbee:
### Resolves

https://github.com/LLK/scratch-render/issues/396, https://github.com/LLK/scratch-vm/issues/1451

### Proposed Changes

Uses an "inset" value in RenderWebGL.js that matches the same logic in [Scratch 2](https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchSprite.as#L209)

### Reason for Changes

`getFencedPositionOfDrawable` in RenderWebGL.js currently sets an inset for [width](https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L1305) and [height](https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L1311) instead of one inset for both like [Scratch 2 uses.](https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchSprite.as#L209) This means that sprites that aren't close to square will land at a different point when set to go off stage. 

Here is an example project that shows that this change successfully allows you to fall into pits. I remixed this project from the original to set the `y position` check in `check edges` for the player sprite back to the original value, -187. (Thanks for recording what it originally was @BryceLTaylor! super helpful once the user changed it).

https://scratch.mit.edu/projects/279811694/

### Test Coverage

[I have created an sb2 test](https://github.com/LLK/scratch-render/files/2765545/sprite-goes-off-stage.sb2.zip) that can be run by scratch-tests.js, but I found that it is broken for more reasons than just the one covered in my PR. In the test I set the sprite to go off the stage in the X or Y position, and then check whether it has the same value as it did in Scratch 2. Currently with this PR's fix, it fails due to an integer vs. float value. The message below is an example:

```
not ok cat lands off stage left. X position should be -272, and it is -271.15799713
```
My question: should I fix this incompatibility in this PR?


ktbee:
@cwillisf I see what you mean about `inset` being handled pretty differently in Scratch 2, so I've updated this PR title and my commit message to make the claim a little less strongly that the logic now exactly matches Scratch 2. This PR only makes it closer :)

I also wondered about the differences in values between Scratch 2 and 3's fence width values. When I switch Scratch 3 to use `18`, the results from my sprite-goes-off-stage.sb2 test are even more wrong, but by varying amounts between -3 and -.5. It looks like `15` is more right for Scratch 3 to be compatible with Scratch 2, though I'm also not sure why.

If this PR looks good to you as a step towards fixing some incompatibility, I can make a follow up issue to investigate the differences in my sprite-goes-off-stage.sb2 test.

Issue #390, opened 2019-01-04
Touching color/sprite inaccurate on scaled bitmaps


cwillisf:
### Expected Behavior

The touching color & touching sprite blocks should match the visible results on the stage.

### Actual Behavior

The touching color & touching sprite blocks can give false results when a scaled-up bitmap is involved.

### Steps to Reproduce

Test project: https://scratch.mit.edu/projects/277169481/

1. Make a small bitmap costume, preferably with transparent gaps between solid parts of the costume.
2. Scale up the sprite with the bitmap costume such that the sprite occupies most of the stage.
3. Make another small sprite and test "touching sprite" and "touching color" against the scaled-up sprite.

### Operating System and Browser

macOS 10.13.6, Chrome 71.0.3578.98

cwillisf:
This might be due to unwanted linear filtering on the query canvas, similar to some of the blurriness issues we've been working on this week.

fsih:
Still trying to understand everything that goes on in the renderer, but I think there are issues in 2 places in RenderWebGL's isTouchingColor

Here:
https://github.com/LLK/scratch-render/blob/a0ce9e3dca18c8419b649948aee624838ea28b8a/src/RenderWebGL.js#L743
and here:
https://github.com/LLK/scratch-render/blob/a0ce9e3dca18c8419b649948aee624838ea28b8a/src/RenderWebGL.js#L745

These functions eventually use Silhouette's isTouchingNearest and colorAtNearest, which sample a pixel off of a super small pixel representation of the tiny vector at "original size".

@cwillisf What is the silhouette meant to represent? Should it be scaled when the sprite is scaled on stage? Or should the touching color and sprite code be trying to sample something other than the silhouette? (you mentioned a query canvas?)

fsih:
@cwillisf thanks for the thorough write-up!
When you say we should build a new silhouette when the SVG skin is re-rasterized, what do you mean by re-rasterized? Do you mean the setSVG call in SVGSkin.js?

PR #404, opened 2019-01-22
Update silhouette after getting texture at a new scale

fsih:
### Resolves

https://github.com/LLK/scratch-render/issues/390

### Proposed Changes

Update the silhouette after redrawing a texture at a different scale

### Reason for Changes
Use a silhouette rendered the same size as what's on stage

### Test Coverage
Tested project 277169481


cwillisf:
It depends on the costume/skin size. In most cases I would expect to see a rescale when going into full-screen mode but maybe not when just toggling non-full-screen stage size. Is that what you were seeing?

If not it could just mean that I'm forgetting something, but even if it's a bug I think it would be an independent issue from this one :)

PR #406, opened 2019-01-28
Add "query playground"

cwillisf:
### Resolves

Progress toward diagnosing #390 and other "touching color" issues

### Proposed Changes

1. Add a flag to force the renderer to use the GPU or CPU implementation of `touching color?`.
2. Pull playground JS out of the HTML and into its own JS files.
3. Build the playground JS using Webpack instead of just copying it as-is.
4. Add a new playground, `queryPlayground.html`, to interactively test renderer queries like `touching color?`.

Current state:
image

### Reason for Changes

This should help figure out what's going wrong with `touching color?`, as well as making it easier to add more playgrounds in the future if necessary.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -132,6 +132,9 @@ class RenderWebGL extends EventEmitter {
             throw new Error('Could not get WebGL context: this browser or environment may not support WebGL.');
         }
 
+        /** @type {RenderWebGL.UseGpuModes} */
+        this._useGpuMode = RenderWebGL.UseGpuModes.Automatic;
+
         /** @type {Drawable[]} */
         this._allDrawables = [];
 
@@ -244,18 +247,11 @@ class RenderWebGL extends EventEmitter {
     }
 
     /**
-     * Force "touching color" operations to use (or not use) the GPU. By default the renderer will decide.
-     * @param {boolean} forceGPU - true to force the renderer to use the GPU, false to force CPU.
+     * Control the use of the GPU or CPU paths in `isTouchingColor`.
+     * @param {RenderWebGL.UseGpuModes} useGpuMode - automatically decide, force CPU, or force GPU.
      */
-    setForceGPU (forceGPU) {
-        this._forceGPU = !!forceGPU;
-    }
-
-    /**
-     * Clear any value set by `setForceGPU`, allowing the renderer to decide.
-     */
-    clearForceGPU () {
-        this._forceGPU = null;
+    setUseGpuMode (useGpuMode) {
+        this._useGpuMode = useGpuMode;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/RenderWebGL.js
```diff
@@ -243,6 +243,21 @@ class RenderWebGL extends EventEmitter {
         this._debugCanvas = canvas;
     }
 
+    /**
+     * Force "touching color" operations to use (or not use) the GPU. By default the renderer will decide.
+     * @param {boolean} forceGPU - true to force the renderer to use the GPU, false to force CPU.
+     */
+    setForceGPU (forceGPU) {
```
mzgoddard: Maybe use an enum instead? One that accounts for the three options? Like `ONLY_GPU` `MIXED_CPU_GPU` `ONLY_CPU`?

src/RenderWebGL.js
```diff
@@ -717,9 +732,18 @@ class RenderWebGL extends EventEmitter {
 
         const bounds = this._candidatesBounds(candidates);
 
-        // if there are just too many pixels to CPU render efficently, we
-        // need to let readPixels happen
-        if (bounds.width * bounds.height * (candidates.length + 1) >= __cpuTouchingColorPixelCount) {
+        const maxPixelsForCPU = (typeof this._forceGPU === 'boolean') ?
```
mzgoddard: Having these two `_forceGPU` tests feels weird to me. Would testing explicitly `=== true`, `=== false`, or `=== null` be better?

src/RenderWebGL.js
```diff
@@ -728,23 +752,27 @@ class RenderWebGL extends EventEmitter {
...
-                ) {
-                    return true;
+                // if we use a mask, check our sample color...
+                if (hasMask ?
+                    maskMatches(Drawable.sampleColor4b(point, drawable, color), mask3b) :
+                    drawable.isTouching(point)) {
+                    RenderWebGL.sampleColor3b(point, candidates, color);
+                    if (debugCanvasContext) {
+                        debugCanvasContext.fillStyle = `rgb(${color[0]},${color[1]},${color[2]})`;
+                        debugCanvasContext.fillRect(x - bounds.left, bounds.bottom - y, 1, 1);
```
mzgoddard: Maybe this is a weird idea. What if this wasn't drawing but was calling delegate methods. Methods on debugCanvasContext. Then the different playground pages could change what they use about this data if that would be useful.

cwillisf: I like the idea of delegates. That could also solve another problem I've been having, where using "the" debug canvas is basically impossible in some projects since you only get a brief glimpse of the thing you're interested in before the canvas gets overwritten by the next thing. This could be solved by including some basic context info when calling the delegate method so the delegate could decide which data to handle, and how.

cwillisf: I explored this idea a bit and it started getting super complicated, mainly due to the differences between the CPU and GPU paths. It might be better (or at least easier) to have canvas-fetching delegates or something like that... I dunno. Anyway, I still like the idea but I think I'm filing it under "future work"...

Issue #214, opened 2017-12-12
Touching color consistency with Scratch 2


paulkaplan:
This is an issue to track example projects and repros of inconsistencies with touching-color between scratch 3 and scratch 2. This was inspired by https://github.com/LLK/scratch-gui/issues/689, which gives the first example case:

- [ ] https://llk.github.io/scratch-gui/develop/#171494108

thisandagain:
Here is a more minimal repro that demonstrates the issue:

https://scratch.mit.edu/projects/193005590/
https://llk.github.io/scratch-gui/develop/#193005590

paulkaplan:
This PR (and comments) provide some pretty good background, @cwillisf has put some good thinking into this issue already https://github.com/LLK/scratch-render/pull/162

LiFaytheGoblin:
I am wondering if the tolerance issue mentioned before is also the reason for this 2.0 project not being compatible: 
Touching color works not when
<img width="667" alt="screen shot 2018-08-29 at 1 51 19 pm" src="https://user-images.githubusercontent.com/5471273/44805744-0d828580-ab93-11e8-86db-7749b64ef2c3.png">
<img width="743" alt="screen shot 2018-08-29 at 1 51 39 pm" src="https://user-images.githubusercontent.com/5471273/44805753-11aea300-ab93-11e8-9b9b-2d6d31e2d3a2.png">
<img width="654" alt="screen shot 2018-08-29 at 1 52 19 pm" src="https://user-images.githubusercontent.com/5471273/44805757-15422a00-ab93-11e8-9ff4-af73f1b170da.png">

But works when
<img width="654" alt="screen shot 2018-08-29 at 1 51 03 pm" src="https://user-images.githubusercontent.com/5471273/44805771-1bd0a180-ab93-11e8-90b1-6ce244fc177b.png">

So, not working for anything smaller than 3x3 pixels and also the positioning of the pixels seem to be important. Found in this project: 64636198

BryceLTaylor:
The thing that's changing in those images and causing a difference is the size of the pink square.  @LiFaytheGoblin was adding and removing pixels in the paint editor to change it's size.

Mr-Dave2:
A color sensing issue was mentioned in the forums:  https://scratch.mit.edu/discuss/topic/308114/?page=14#post-3339707 
ToadfanSchool submitted this as a incompatible project

https://scratch.mit.edu/projects/202141892/
https://beta.scratch.mit.edu/#202141892
[...]

PR #407, opened 2019-02-04
Adjust CPU isTouchingColor to match GPU results

cwillisf:
### Resolves

Fixes some causes of #214

### Proposed Changes

Functionality changes:
- Adjust the loops in `isTouchingColor` to avoid testing extra edge pixels
- Adjust the math in `colorAtNearest` to match hardware texture sampling results
- `sampleColor4b` no longer contains an early-out bounds test since `colorAtNearest` does the same test more accurately. For example, we were previously getting incorrect results when "zero" turned into a very small negative number due to floating point error.

Other changes:
- Move projection matrix calculation into a separate function
   - An earlier version of this code contained some adjustments to the projection matrix. It turns out those adjustments weren't needed, but I decided to keep the centralized function anyway.
- Spelling corrections in comments

### Reason for Changes

The CPU version of `isTouchingColor`, `sampleColor4b`, and a few related functions were not returning the same results as their GPU equivalents. This could cause unexpected behavior, especially if a project uses both paths.

### Test Coverage

See #406

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -747,8 +747,8 @@ class RenderWebGL extends EventEmitter {
                 return this._isTouchingColorGpuFin(bounds, color3b, y);
             }
             for (let x = 0; x < bounds.width; ++x) {
-                point[1] = bounds.top - y; // bounds.top <= y < bounds.bottom ("flipped")
-                point[0] = bounds.left + x; // bounds.left <= x < bounds.right
+                point[0] = bounds.left + x; // bounds.left <= point[0] < bounds.right
+                point[1] = bounds.top - y; // bounds.bottom < point[1] <= bounds.top ("flipped")
                 // if we use a mask, check our sample color...
                 if (hasMask ?
                     maskMatches(Drawable.sampleColor4b(point, drawable, color), mask3b) :
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -256,8 +256,7 @@ class RenderWebGL extends EventEmitter {
         this._yBottom = yBottom;
         this._yTop = yTop;
 
-        // swap yBottom & yTop to fit Scratch convention of +y=up
-        this._projection = twgl.m4.ortho(xLeft, xRight, yBottom, yTop, -1, 1);
+        this._projection = this._makeOrthoProjection(xLeft, xRight, yBottom, yTop);
 
         this._setNativeSize(Math.abs(xRight - xLeft), Math.abs(yBottom - yTop));
     }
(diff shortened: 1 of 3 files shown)


src/Silhouette.js
```diff
@@ -120,12 +120,10 @@ class Silhouette {
      * @returns {Uint8ClampedArray} dst
      */
     colorAtNearest (vec, dst) {
-        return getColor4b(
-            this,
-            Math.floor(vec[0] * (this._width - 1)),
-            Math.floor(vec[1] * (this._height - 1)),
-            dst
-        );
+        const x = Math.round(vec[0] * this._width);
```
gnarf: I went back and forth a long time on if -1 was needed here or not -- do you have some ninja unit/integration testing that proves it isn't?

src/RenderWebGL.js
```diff
@@ -728,23 +741,23 @@ class RenderWebGL extends EventEmitter {
...
-                    // if we use a mask, check our sample color
-                    (hasMask ?
-                        maskMatches(Drawable.sampleColor4b(point, drawable, color), mask3b) :
-                        drawable.isTouching(point)) &&
-                    // and the target color is drawn at this pixel
-                    colorMatches(RenderWebGL.sampleColor3b(point, candidates, color), color3b, 0)
-                ) {
-                    return true;
+            for (let x = 0; x < bounds.width; ++x) {
+                point[1] = bounds.top - y; // bounds.top <= y < bounds.bottom ("flipped")
```
gnarf: I really don't like the fact that this ends up with `y !== point[1]` -- can we maybe rename the outer loop var to `yFromTop` or something

src/Silhouette.js
cwillisf: I was using #406 extensively while working on this, and skipping the -1 here is one of the main changes that makes the results exactly the same as the GPU version. In fact, this work is why I did #406 in the first place :)

I also drew it out on paper to convince myself that it's correct: take a look at which texel you "want" to sample in a 4-texel-wide image when your texture coordinate is 0.01, 0.24, 0.26, 0.49, 0.51, 0.74, 0.76, and 0.99.

src/RenderWebGL.js
cwillisf: Hmm, I kind of agree but on the other hand if you combine this with #406 then you end up with `debugCanvasContext.fillRect(x, y, 1, 1)` being the correct way to draw to the debug canvas to match the GPU path's results. I'm not saying that clean debug code should be the priority (ha!) but rather I feel like that's a sign that these are the "natural" values for `x` and `y`. (Side note: it's very satisfying to see the CPU path early-out when a matching color is found, leaving the rest of the canvas blank!)

Consider also that `0 <= y < bounds.height` and `0 <= x < bounds.width` are both correct here, as is `bounds.left <= point[0] < bounds.right`, but `bounds.bottom <= point[1] < bounds.top` is wrong.

Actually, that makes me realize that this comment refers to the wrong variable. I'll fix that :/

Issue #403, opened 2019-01-17
X and Y positions differ at edge of stage between Scratch 2 and 3


ktbee:
### Expected Behavior

When you send a sprite off stage (say X position = 1000 and Y = 0), its position should be limited to the edge of the stage and its X and Y values should be the same in Scratch 2 and 3. 

### Actual Behavior

The X or Y value that goes off stage is slightly different in Scratch 3 vs. Scratch 2. In Scratch 2, the X and Y values are integers. In Scratch 3, they are floating point numbers that are within about .5 pixels of the Scratch 2 position value. 

### Steps to Reproduce

I've created an sb2 project as a potential [`scratch-test`](https://github.com/LLK/scratch-render/tree/develop/test/integration/scratch-tests) that reproduces this issue well. You can find this file below:

[sprite-goes-off-stage.sb2.zip](https://github.com/LLK/scratch-render/files/2770334/sprite-goes-off-stage.sb2.zip)

The repro steps:
- In the Scratch 2 offline editor, send a sprite off stage and use the "say" looks block to display the "x position" motion block or the "y position" motion block.
- Save the sb2 project and load it in Scratch 3
- Observe that the X or Y value for when the sprite tries to go off stage (for example go to 1000,0 or to 0,1000) is different than the values for Scratch 2. 

### Operating System and Browser

Mac OS 10.14 Chrome 71.0.3578.98

ktbee:
Looks like there are a couple of factors behind why there's incompatibility between Scratch 2 and 3 for where a sprite is position when it tries to go off stage. I addressed one of the more straight forward ones in this PR, where [we need to use Math.ceil and Math.floor](https://github.com/LLK/scratch-render/pull/408).

[...]

PR #409, opened 2019-02-05
Compatibility fix for bitmap position off stage

ktbee:
### Resolves

Fixes some of the factors in #403 and depends on https://github.com/LLK/scratch-render/pull/408 

### Proposed Changes

Add a method for identifying skin type and if the drawable is a bitmap, always use `getAABB` for its bounding box when calculating its fenced position at the edge of the stage. 

### Reason for Changes

[In Scratch 2, we used `getRect` to get the aabb](https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchSprite.as#L199), whose implementation seems closer to `drawable.getAABB` than `drawable.getBounds`. `drawable.getAABB` seems closer because [Flash's `getRect` doesn't take into account any strokes on the shape,](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/DisplayObject.html#getBounds()) and `drawable.getBounds` uses hull points, which would take stroke into account.  If the skin is a bitmap, we make sure we always use `drawable.getAABB` instead of `drawable.getFastBounds`, which might use either `drawable.getBounds` or `drawable.getAABB`.

All this being said, this fix doesn't work for SVG skins, so we're sticking with our previous tactic for getting the aabb for any non-bitmap skins.

### Test Coverage

I updated sprite-goes-off-stage.sb2 to include more bitmap sprite examples to test.


src/RenderWebGL.js
```diff
@@ -1316,21 +1316,26 @@ class RenderWebGL extends EventEmitter {
...
         const dy = y - drawable._position[1];
+        let aabb;
 
-        const aabb = drawable.getFastBounds();
-        const inset = Math.floor(Math.min(aabb.width, aabb.height) / 2);
+        if (drawable._skin.typeOfSkin === 'BITMAP') {
+            aabb = drawable.getAABB();
+        } else {
+            aabb = drawable.getFastBounds();
+        }
```
cwillisf: I'd prefer to avoid making decisions externally based on types. How would you feel about delegating this to a method on the skin? For example, `getFenceBounds` could call (or even _be_) `getAABB` for bitmaps and `getFastBounds` elsewhere.

ktbee: @cwillisf that makes sense to me! I'll make that change now

ktbee:
Thank you for the suggestion/feedback @cwillisf!  I've implemented it in my latest round of changes, let me know what you think.

PR #410, opened 2019-02-07
Revert "Adjust CPU isTouchingColor to match GPU results"

cwillisf:
Reverts LLK/scratch-render#407

It appears this change causes thin lines to be missed sometimes. To be investigated:
image


cwillisf:
Project to reproduce issue:
[touching color_.sb3.zip](https://github.com/LLK/scratch-render/files/2877666/touching.color_.sb3.zip)

PR #414, opened 2019-02-14
ImageData WebGL Textures

mzgoddard:
### Resolves

Load faster and use less memory.

### Proposed Changes

- Support ImageData as an argument to Silhouette.update
- Get image data out of input canvases to BitmapSkin.setBitmap and SVGSkin.setSVG. And pass image data to webgl when creating or updating the texture.

### Reason for Changes

Getting the ImageData in setBitmap and setSVG lets us skip an extra draw operation to Silhouette's update canvas. It likely also skips a step browsers perform to do a operation similar to getImageData and instead use ImageData directly. This provides a significant reduction in the time it takes to turn asset data in Storage into a Skin.

### Benchmark Data

Pending ...

### DIFF
--- a/src/BitmapSkin.js
+++ b/src/BitmapSkin.js
@@ -79,20 +79,31 @@ class BitmapSkin extends Skin {
     setBitmap (bitmapData, costumeResolution, rotationCenter) {
         const gl = this._renderer.gl;
 
+        // Preferably bitmapData is ImageData. ImageData speeds up updating
+        // Silhouette and is better handled by more browsers in regards to
+        // memory.
+        let textureData = bitmapData;
+        if (bitmapData instanceof HTMLCanvasElement) {
+            // Given a HTMLCanvasElement get the image data to pass to webgl and
+            // Silhouette.
+            const context = bitmapData.getContext('2d');
+            textureData = context.getImageData(0, 0, bitmapData.width, bitmapData.height);
+        }
+
         if (this._texture) {
             gl.bindTexture(gl.TEXTURE_2D, this._texture);
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, bitmapData);
-            this._silhouette.update(bitmapData);
+            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, textureData);
+            this._silhouette.update(textureData);
         } else {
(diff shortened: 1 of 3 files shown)


thisandagain:
Thanks @mzgoddard !

/cc @kchadha @BryceLTaylor

src/BitmapSkin.js
```diff
@@ -79,20 +79,34 @@ class BitmapSkin extends Skin {
...
-                src: bitmapData
+                src: textureData
             };
 
             this._texture = twgl.createTexture(gl, textureOptions);
-            this._silhouette.update(bitmapData);
+            this._silhouette.update(textureData);
+
+            // Unset src to make garbage collecting textureData easier.
+            textureOptions.src = null;
```
cwillisf: Is this really necessary, especially since `textureOptions` is about to fall out of scope? Also, are you certain that it's safe, in that there's no way `twgl` will end up looking at `textureOptions.src` after some async operation?

src/SVGSkin.js
```diff
@@ -99,20 +103,31 @@ class SVGSkin extends Skin {
...
-                    src: this._svgRenderer.canvas
+                    src: textureData
                 };
 
                 this._texture = twgl.createTexture(gl, textureOptions);
-                this._silhouette.update(this._svgRenderer.canvas);
+                this._silhouette.update(textureData);
+
+                // Unset src to make garbage collecting textureData easier.
+                textureData.src = null;
```
cwillisf: (same questions as above)

src/Silhouette.js
```diff
@@ -88,17 +88,27 @@ class Silhouette {
...
-        if (!(width && height)) {
-            return;
+        let imageData;
+        if (bitmapData instanceof ImageData) {
+            // If handed ImageData directly, use it directly.
+            imageData = bitmapData;
+            this._width = bitmapData.width;
+            this._height = bitmapData.height;
+        } else {
+            // Draw about anything else to our update canvas and poll image data
```
cwillisf: `s/poll/pull/`?

mzgoddard: Poll.

I guess I view getImageData like a file system operation.

src/BitmapSkin.js
mzgoddard: Sorry. I really let responding to this slip.

Mostly this is a line from an early iteration of the ImageData change. From before I focused on using ImageData more and was broadly experimenting.

The benefit of this line, which is definitely hard to measure, isn't enough to stall the rest of this PR. I'm going to remove it and the other.

PR #418, opened 2019-02-27
pen transparency fix

peabrainiac:
Clears the pen framebuffer with transparent black instead of transparent white so transparent parts of the pen layer don't show up lighter than they should anymore.

See https://scratch.mit.edu/projects/277120576/.

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -599,7 +599,7 @@ class PenSkin extends Skin {
             this._silhouetteBuffer = twgl.createFramebufferInfo(gl, [{format: gl.RGBA}], width, height);
         }
 
-        gl.clearColor(1, 1, 1, 0);
+        gl.clearColor(0, 0, 0, 0);
         gl.clear(gl.COLOR_BUFFER_BIT);
 
         this._silhouetteDirty = true;
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -155,6 +155,13 @@ class PenSkin extends Skin {
     }
 
     /**
+     * @returns {boolean} true if alpha is premultiplied, false otherwise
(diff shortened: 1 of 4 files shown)


peabrainiac:
> Thanks for tracking down and fixing this!
> 
> I think the other call to `gl.clearColor`, down in `_setCanvasSize`, should be changed too. Even better: you could set a constant near the top of the file and use that constant in both `gl.clearColor` calls.

Yeah, I made the pull request a little too early and forgot to double-check that. There's also one more thing that needs to be changed - I think the pen layer is rendered with the wrong blend functions. However, I couldn't find out which function is responsible for rendering the pen layer to the stage - could you help me out there? @cwillisf

peabrainiac:
Transparent pen now works just as expected, but now there's an issue when stamping transparent sprites. Once that is fixed this should all be working though.

peabrainiac:
Transparency works correctly now, both for pen, stamping and transparent sprites. Looks like the `<is touching color [ ]>`-block is broken though. I'll see if I can fix this.

peabrainiac:
I just tested it on scratch.mit.edu, and it looks like the <is touching color [ ]>-block is broken there too, so it's probably not a problem with my changes here. I think this should be ready for merging now @cwillisf or is there something I still need to change?

thisandagain:
@cwillisf Can you re-review?

cwillisf:
Testing notes: this project should look the same in both Scratch 2.0 and Scratch 3.0. It's based on project 277120576 linked above, but this version loads in the old editor as well.

[pen-blend.sb2.zip](https://github.com/LLK/scratch-render/files/3129546/pen-blend.sb2.zip)

Issue #374, opened 2018-12-05
Re-rasterization of a vector sprite at a higher resolution causes all others with the same costume to change to the high-resolution rasterization


adroitwhiz:
### Expected Behavior

A vector sprite's rendered appearance should be independent of the state of other instances of the sprite.

### Actual Behavior

When a vector sprite reaches a size that triggers its current costume's re-rasterization, other instances with the same costume (e.g. clones) will suddenly switch over to the re-rasterized costume regardless of _their_ size.

### Steps to Reproduce

- 1. Create a clone of a vector sprite.
- 2. Use the "set size" or "change size" blocks on one of the sprite instances, leaving the other at its original size.
- 3. When the resized instance's costume is re-rasterized, note the pixelation that occurs in the unchanged instance's costume.

You can test this with [this project](https://github.com/LLK/scratch-render/files/2649688/vector.sizing.demo.zip)-- just change the extension to a .sb3, click the green flag, and then press space. [In addition, here's a video of the bug.](https://youtu.be/0fug-56tLDA)

### Operating System and Browser

Fedora Workstation 29 x86_64, Chromium 70.0.3538.110

ktbee:
After taking a look at this issue and chatting with @mzgoddard, it looks like this low resolution for the clone skin has the same root cause as https://github.com/LLK/scratch-render/issues/334. The skin is scaled up for the original sprite, but we don't currently have logic to scale down the same skin for the clone, so it's not shown at its original resolution. 

[As we were discussing in this and the following comments](https://github.com/LLK/scratch-render/issues/334#issuecomment-449143929), we could implement mipmaps/different skin resolutions using JavaScript or WebGL. Any thoughts on this now that we're revisiting it, @cwillisf ?

Issue #334, opened 2018-08-20
 Poor quality of vector images in full screen mode


thisandagain:
### Expected Behavior

Good anti-aliasing

### Actual Behavior

Incorrect anti-aliasing

### Steps to Reproduce

1.Open Scratch 3.0
2.Click on full screen

![x](https://user-images.githubusercontent.com/28606130/43994373-8e3fda1c-9da4-11e8-8682-8419c601f39f.png)

### Operating System and Browser

Windows 10, Chrome v.68.0.3440.106

---
Moved from LLK/scratch-gui#2847
/cc @SimaKyr

thisandagain:
@cwillisf This is one of the top pieces of feedback we are seeing in the forums. Let's discuss next week. 

/cc @fsih @kchadha @BryceLTaylor @rschamp

BryceLTaylor:
Her's a great repro of this from the community:
https://llk.github.io/scratch-gui/develop/compatibility-testing.html#230352473

It stamps a white sprite on a white background.  I bet it's not actually a problem with the stamping itself, but with the image itself.

fsih:
@BryceLTaylor I think that might be https://github.com/LLK/scratch-render/issues/237 , not sure if it's the same issue

BryceLTaylor:
@fsih It does seem that it is, I didn't know about the other issue.  They do seem related in that they have to do with aliasing.

adroitwhiz:
I really hope this will be fixed before release. I hopped on Scratch 2.0 as soon as the public beta was available, but the terrible vector rendering artifacts (along with the block size issue) are why I haven't done the same with 3.0 and am hanging on to 2.0 for dear life.

Vector sprite quality as it stands is unacceptable. Between this bug, #246, #237, and #233, the vector rendering pipeline feels worse than half-baked and in desperate need of an overhaul in my opinion.

ktbee:
Taking a look at this issue, it looks like currently [`getTexture`](https://github.com/LLK/scratch-render/blob/develop/src/SVGSkin.js#L68-L88) only scales textures up and doesn't have logic to scale textures back down again. What would y'all think about an approach where we instead generate textures at whichever scale we need (the first one starting at the sprite's saved size) and save them in an array in case we need them for later use?

Also aside from  `getTexture`, it looks like another codepath might be at play for when the texture is scaled up. I think we can use this same array of different sized textures to solve that issue as well.

/cc @fsih @cwillisf

[...]

PR #431, opened 2019-04-01
Add JavaScript MIPs for scaling textures larger and smaller

ktbee:
### Resolves

https://github.com/LLK/scratch-render/issues/374, potentially also https://github.com/LLK/scratch-render/issues/334

### Proposed Changes

[As discussed on other issues](https://github.com/LLK/scratch-render/issues/334), adding mipmap functionality will be helpful for scaling images at a good resolution in Scratch. This PR:

- Adds a SVGMIP as a wrapper for WebGL textures to handle creating them at a certain scale and updating them if needed
- Stores JavaScript mips on SVGSkin in an array `_scaledMIPs` using an index offset to allow for mips that have a scale of less than one. We're using an array here for better performance instead of an object with more descriptive keys
- Extends while loop in `SVGSkin.getTexture` to get smaller scales in addition to larger ones

Here is an example repro project. After the green flag click, the original sprite is scaled up and down while the clone remains unscaled. Both have a better resolution because they no longer share a texture at all times and use the one at its requested scale. https://scratch.mit.edu/projects/295219677/editor/

The screenshots' differences are a little tough to see at a smaller size, but it's clear when you click them and see them in a new tab. Also these were taken on a lower resolution monitor. If your screen has a hidpi, these differences are harder to see. 

| Production     | PR Changes |
| ----------- | ----------- |
| Screen Shot 2019-04-01 at 2 28 15 PM |  Screen Shot 2019-04-01 at 2 25 26 PM |

### Memory usage

[In our previous discussion of this topic,](https://github.com/LLK/scratch-render/issues/334#issuecomment-449143929) one point against scaling mips with JavaScript instead of automatically on the GPU is that this method would use more memory. From some profiling I did with [Chrome](https://www.lifewire.com/google-chrome-task-manager-4103619) and [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Performance/about:memory), it seems like [this test project](https://scratch.mit.edu/projects/295219677/editor/) hovers around the same amount of memory on the `develop` branch and this feature branch. Here are the results I found:

| **Chrome**  | `develop`      | `javascript-scaled-textures` |
| ----------- | ----------- | ----------- |
| **Memory Footprint** | 192 MB |  190 MB  |
| **GPU Memory**  |  28.3 MB      |    28.5 MB |

| **Firefox**  | `develop`      | `javascript-scaled-textures` |
| ----------- | ----------- | ----------- |
|Trial 1  |  81.96 MB (09.74%)  | 82.49 MB (09.79%) |
|Trial 2  | 73.19 MB (08.73%)  | 104.62 MB (11.86%) |
|Trial 3  | 99.08 MB (11.43%) | 84.35 MB (09.85%) |

Theoretically there should be a memory usage increase because there are more textures. I added the surface areas of power of 2 sized images between 2x2 and 1024x1024 and found that they would use about 33% more memory than just a single 1024x1024 image, but perhaps in the way we're using them in this branch, the browser is able to release some of that memory. From what I can see using Chrome and Firefox's tools, multiple JavaScript scaled textures hasn't had much of an impact on a tab's memory usage. I'm curious to hear more thoughts on this though!

### What this hasn't solved yet

I've found this implementation of mips successfully scales sprites that have been changes by a "looks" block, but sprites that haven't been scaled by a block still don't scale well. For example, when you reduce the browser's width while using the editor, the stage scales down and the sprite's quality is reduced:

| Full size stage      | Smaller stage |
| -----------            | ----------- |
| Screen Shot 2019-04-01 at 2 17 11 PM     | Screen Shot 2019-04-01 at 2 17 23 PM |

Fullscreen mode is another example of this:

Screen Shot 2019-04-01 at 2 32 13 PM

Thoughts on how to scale sprites in this scenario and also thoughts on this approach for mipmaps are much appreciated!

### DIFF
--- a/src/SVGMIP.js
+++ b/src/SVGMIP.js
@@ -1,0 +1,56 @@
+const twgl = require('twgl.js');
+
+class SVGMIP {
+    /**
+    * Create a new SVG MIP for a given scale.
+    * @param {RenderWebGL} renderer - The renderer which this MIP's skin uses.
+    * @param {SvgRenderer} svgRenderer - The svg renderer which this MIP's skin uses.
+    * @param {number} scale - The relative size of the MIP
+    * @param {function} callback - A callback that should always fire after draw()
+    * @constructor
+    */
+    constructor (renderer, svgRenderer, scale, callback) {
+        this._renderer = renderer;
+        this._svgRenderer = svgRenderer;
+        this._scale = scale;
+        this._texture = null;
+        this._callback = callback;
+
+        this.draw();
+    }
+
+    draw () {
+        this._svgRenderer._draw(this._scale, () => {
+            const textureData = this._getTextureData();
+            const textureOptions = {
+                auto: false,
+                wrap: this._renderer.gl.CLAMP_TO_EDGE,
+                src: textureData
+            };
+
+            this._texture = twgl.createTexture(this._renderer.gl, textureOptions);
+            this._callback(textureData);
+        });
+    }
+
+    dispose () {
+        this._renderer.gl.deleteTexture(this.getTexture());
+    }
+
+    getTexture () {
+        return this._texture;
+    }
+
+    _getTextureData () {
+        // Pull out the ImageData from the canvas. ImageData speeds up
+        // updating Silhouette and is better handled by more browsers in
+        // regards to memory.
+        const canvas = this._svgRenderer.canvas;
+        const context = canvas.getContext('2d');
+        const textureData = context.getImageData(0, 0, canvas.width, canvas.height);
+
+        return textureData;
+    }
+}
+
+module.exports = SVGMIP;
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -1,9 +1,16 @@
-const twgl = require('twgl.js');
-
 const Skin = require('./Skin');
+const SVGMIP = require('./SVGMIP');
 const SvgRenderer = require('scratch-svg-renderer').SVGRenderer;
 
 const MAX_TEXTURE_DIMENSION = 2048;
+const MIN_TEXTURE_SCALE = 1 / 256;
+/**
+ * All scaled renderings of the SVG are stored in an array. The 1.0 scale of
+ * the SVG is stored at the 8th index. The smallest possible 1 / 256 scale
+ * rendering is stored at the 0th index.
+ * @const {number}
+ */
+const INDEX_OFFSET = 8;
 
 class SVGSkin extends Skin {
     /**
@@ -25,11 +32,14 @@ class SVGSkin extends Skin {
         /** @type {WebGLTexture} */
         this._texture = null;
 
-        /** @type {number} */
-        this._textureScale = 1;
+        /** @type {Array.<SVGMIPs>} */
+        this._scaledMIPs = [];
 
-        /** @type {Number} */
-        this._maxTextureScale = 0;
+        /**
+        * Ratio of the size of the SVG and the max size of the WebGL texture
+        * @type {Number}
+        */
+        this._maxTextureScale = 1;
     }
 
     /**
@@ -37,8 +47,13 @@ class SVGSkin extends Skin {
      */
     dispose () {
         if (this._texture) {
-            this._renderer.gl.deleteTexture(this._texture);
+            for (const mip of this._scaledMIPs) {
+                if (mip) {
+                    mip.dispose();
+                }
+            }
             this._texture = null;
+            this._scaledMIPs.length = 0;
         }
         super.dispose();
     }
@@ -61,10 +76,32 @@ class SVGSkin extends Skin {
     }
 
     /**
+     * Create a MIP for a given scale and pass it a callback for updating
+     * state when switching between scales and MIPs.
+     * @param {number} scale - The relative size of the MIP
+     * @param {function} resetCallback - this is a callback for doing a hard reset
+     * of MIPs and a reset of the rotation center. Only passed in if the MIP scale is 1.
+     * @return {SVGMIP} An object that handles creating and updating SVG textures.
+     */
+    createMIP (scale, resetCallback) {
+        const textureCallback = textureData => {
+            if (resetCallback) resetCallback();
+            // Check if we have the largest MIP
+            // eslint-disable-next-line no-use-before-define
+            if (!this._scaledMIPs.length || this._scaledMIPs[this._scaledMIPs.length - 1]._scale <= scale) {
+                // Currently silhouette only gets scaled up
+                this._silhouette.update(textureData);
+            }
+        };
+        const mip = new SVGMIP(this._renderer, this._svgRenderer, scale, textureCallback);
+
+        return mip;
+    }
+
+    /**
      * @param {Array<number>} scale - The scaling factors to be used, each in the [0,100] range.
      * @return {WebGLTexture} The GL texture representation of this skin when drawing at the given scale.
      */
-    // eslint-disable-next-line no-unused-vars
     getTexture (scale) {
         if (!this._svgRenderer.canvas.width || !this._svgRenderer.canvas.height) {
             return super.getTexture();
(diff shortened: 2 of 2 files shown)


adroitwhiz:
> Did you consider using OpenGL's built-in mipmap support? There might be a reason not to choose this that I haven't thought of, but it seems like it would simplify the code.

IIRC, OpenGL mipmaps only work with textures whose dimensions are a power of 2.

ktbee:
We discussed this a bit offline, but just to follow up for anyone watching the issue: 

@adroitwhiz has a good point, for WebGL 1, automatic mipmaps only work for images with dimensions that are a power of two. [We'd need to use an OpenGL extension or upgrade our WebGL version to be able to automatically generate mipmaps for non-power of two images.](https://www.khronos.org/webgl/wiki/WebGL_and_OpenGL_Differences#Non-Power_of_Two_Texture_Support)

ktbee:
Also @cwillisf @kchadha, I turned off twgl's automatic mipmap creation option in this PR and it is all set to be reviewed!

ktbee:
Hooray for tests! They helped me catch a couple bugs, which I fixed with the following changes:

- Use `this._texture.getTexture()` instead of just `this._texture` in SVGSkin's dispose function.
- Updated the silhouette for sprites with a scale of 1 to make sure it's set initially even if the sprite isn't scaled. 

@cwillisf let me know what you think!

adroitwhiz:
Shouldn't `SVGSkin.dispose` dispose of *all* `SVGMIP` textures, instead of just the current one?

src/SVGSkin.js
```diff
@@ -25,19 +26,22 @@ class SVGSkin extends Skin {
...
+        this._largestTextureScale = 1;
     }
 
     /**
      * Dispose of this object. Do not use it after calling this method.
      */
     dispose () {
         if (this._texture) {
-            this._renderer.gl.deleteTexture(this._texture);
+            this._renderer.gl.deleteTexture(this._texture.getTexture());
```
adroitwhiz: Shouldn't `SVGSkin.dispose` call `gl.deleteTexture` on *all* `SVGMIP` textures, instead of just the current one?

ktbee: Oh good point! I've updated `dispose` to delete all of the textures and reset `_scaledMIPs` and `_largestTextureScale`. Let me know if I'm missing anything else.

src/SVGSkin.js
```diff
@@ -1,9 +1,10 @@
-const twgl = require('twgl.js');
-
 const Skin = require('./Skin');
+const SVGMIP = require('./SVGMIP');
 const SvgRenderer = require('scratch-svg-renderer').SVGRenderer;
 
 const MAX_TEXTURE_DIMENSION = 2048;
```
adroitwhiz: Could `gl.getParameter(gl.MAX_TEXTURE_SIZE)` potentially be used for this?

src/SVGSkin.js
```diff
@@ -60,6 +70,34 @@ class SVGSkin extends Skin {
...
+     */
+    createMIP (scale) {
+        const callback = textureData => {
+            if (scale > this._largestTextureScale) {
+                this._largestTextureScale = scale;
+                // Currently silhouette only gets scaled up
+                this._silhouette.update(textureData);
+            }
+
+            if (scale === 1) {
```
adroitwhiz: This seems a bit flaky-- it looks like it assumes `createMIP` will first be called with scale 1, which may not be the case. Could this be moved elsewhere, perhaps to `setSVG`?

ktbee: Hmm, it's not meant to assume `createMIP` will be called with a scale 1. The intention instead is _if_ it gets called with a scale of 1, use the `textureData` to set the `_maxTextureScale` and update the silhouette. We need it to be in the callback in `createMIP` instead of `setSVG` because we need the `textureData` that is passed to the callback in `SVGMIP.draw`.

fsih: I did this sequence of steps and this one weird trick happened
line cat

fsih: This thing I saw may have been https://github.com/LLK/scratch-gui/issues/5215

src/SVGSkin.js
```diff
@@ -1,9 +1,10 @@
-const twgl = require('twgl.js');
-
 const Skin = require('./Skin');
+const SVGMIP = require('./SVGMIP');
 const SvgRenderer = require('scratch-svg-renderer').SVGRenderer;
 
 const MAX_TEXTURE_DIMENSION = 2048;
+const MIN_TEXTURE_SCALE = 1 / 256;
+const INDEX_OFFSET = 8;
```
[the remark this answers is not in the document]
mzgoddard: How about:

```js
/**
 * All scaled renderings of the SVG are stored in an array. The 1.0 scale of
 * the SVG is stored at the 8th index. The smallest possible 1 / 256 scale
 * rendering is stored at the 0th index.
 * @const {number}
 */
```

src/SVGSkin.js
```diff
@@ -25,20 +26,29 @@ class SVGSkin extends Skin {
...
-        this._textureScale = 1;
+        /** @type {Array.<SVGMIPs>} */
+        this._scaledMIPs = [];
+
+        /** @type {Number} */
+        this._maxTextureScale = 1;
 
         /** @type {Number} */
-        this._maxTextureScale = 0;
+        this._largestTextureScale = 1;
```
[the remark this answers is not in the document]
mzgoddard: Combined with another comment, we could get rid of _largestTextureScale. It is needed if we keep already created MIPs. If we dump all MIPs on setSVG calls, we don't need _largestTextureScale.

src/SVGSkin.js
```diff
@@ -100,43 +137,25 @@ class SVGSkin extends Skin {
...
+        } else {
+            const textureScale = 1;
+            this._texture = this._scaledMIPs[INDEX_OFFSET] = this.createMIP(textureScale);
+        }
 
-            if (typeof rotationCenter === 'undefined') rotationCenter = this.calculateRotationCenter();
-            this.setRotationCenter.apply(this, rotationCenter);
-            this.emit(Skin.Events.WasAltered);
-        });
+        if (typeof rotationCenter === 'undefined') rotationCenter = this.calculateRotationCenter();
```
mzgoddard: In a different change we realized that setting the rotationCenter before the image has drawn will cause a "flash" of the old image at the wrong scale. We need to setRotationCenter after drawing.

One way we might do that is move the rotationCenter set logic in createMIP?

src/SVGSkin.js
```diff
@@ -100,43 +137,25 @@ class SVGSkin extends Skin {
...
-                const textureOptions = {
-                    auto: true,
-                    wrap: gl.CLAMP_TO_EDGE,
-                    src: textureData
-                };
-
-                this._texture = twgl.createTexture(gl, textureOptions);
-                this._silhouette.update(textureData);
-            }
+        this._svgRenderer.loadString(svgData);
```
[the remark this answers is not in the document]
mzgoddard: We could:
1. Call dispose on each current MIPs.
2. Create a new array.
3. Set the 1.0 scale MIP at INDEX_OFFSET.

src/SVGMIP.js
```diff
@@ -0,0 +1,69 @@
...
+    */
+    constructor (renderer, svgRenderer, scale, callback) {
+        this._renderer = renderer;
+        this._svgRenderer = svgRenderer;
+        this._scale = scale;
+        this._texture = null;
+        this._callback = callback;
+        this.dirty = false;
+
+        this.draw();
```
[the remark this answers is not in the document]
mzgoddard: It doesn't have to be for every MIP. But we should at least call draw in setSVG on the 1.0 scale MIP. This "primes" the SVGRenderer by loading an Image instance and drawing it to a canvas. The loading part can need a lot of time, so we should load the svg during the initial project load to avoid potentially large draw times when you switch to a SVG costume that hasn't been drawn before.

src/SVGSkin.js
```diff
@@ -60,11 +75,34 @@ class SVGSkin extends Skin {
...
+    createMIP (scale, resetCallback) {
+        const textureCallback = textureData => {
+            // Check if we have the largest MIP
+            // eslint-disable-next-line no-use-before-define
+            if (!this._scaledMIPs.length || this._scaledMIPs[this._scaledMIPs.length - 1]._scale <= scale) {
+                // Currently silhouette only gets scaled up
+                this._silhouette.update(textureData);
+            }
+
+            if (resetCallback) resetCallback();
```
mzgoddard: Run resetCallback before checking scaledMIPs.length. As is, this is a race condition. If you change the svgData, the second time, scaledMIPs may have an index higher than 8 (the 1.0 index) set. During that second change the silhouette will not update for the 1.0 scale MIP.

src/SVGSkin.js
```diff
@@ -60,11 +75,34 @@ class SVGSkin extends Skin {
...
+     * @param {number} scale - The relative size of the MIP
+     * @param {function} resetCallback - this is a callback for doing a hard reset
+     * of MIPs and a reset of the rotation center. Only passed in if the MIP scale is 1.
+     * @return {SVGMIP} An object that handles creating and updating SVG textures.
+     */
+    createMIP (scale, resetCallback) {
+        const textureCallback = textureData => {
+            // Check if we have the largest MIP
+            // eslint-disable-next-line no-use-before-define
+            if (!this._scaledMIPs.length || this._scaledMIPs[this._scaledMIPs.length - 1]._scale <= scale) {
```
mzgoddard: I think we can make this a little simpler.

One way would be if you pass another argument to createMIP. Pass the scaled index and we can `this._scaledMIPs.length - 1 === scaleIndex` to test for this.

src/SVGSkin.js
```diff
@@ -73,80 +111,70 @@ class SVGSkin extends Skin {
...
-            // Pull out the ImageData from the canvas. ImageData speeds up
-            // updating Silhouette and is better handled by more browsers in
-            // regards to memory.
-            const canvas = this._svgRenderer.canvas;
-
-            if (!canvas.width || !canvas.height) {
-                super.setEmptyImageData();
-                return;
-            }
+    resetMIPs (mip, rotationCenter) {
```
mzgoddard: I think this method is pretty slick.

src/SVGSkin.js
ktbee: Makes sense! Thanks for catching that.

BryceLTaylor:
Test Plan:

Description: We will now take vector images and render them at several different sizes in memory to make scaling images look better.

To test:
* clone the sprite so you can see it and the original and try all of the below (make changes to both)
* Take a vector image and scale it to different sizes
* Make sure it looks right when making it bigger
* Make sure it looks right when making it smaller
* Start with a huge sprite and make it very small
* Start with a tiny sprite and make it tinier
* Start with a very small sprite and make it very big
* Edit a sprite while it is in different sizes
* Make the bounds of the sprite larger or smaller (erase parts)
* add graphics effects at different sizes
* rotate the sprite at different sizes

It's important to get Browser coverage for this issue

PR #443, opened 2019-04-20
Always use linear interpolation on vector skins with distortion effects applied

adroitwhiz:
### Resolves

There's no issue for it, but currently, vector sprites with distortion effects (fisheye, whirl, pixelate, and mosaic) appear jagged near 100% scale but smooth otherwise, as can be seen in [this test project](https://scratch.mit.edu/projects/303721927/):
jaggies

### Proposed Changes

This change adds another check to ```Drawable.useNearest()``` so that it returns ```false``` on non-bitmap skins with fisheye, whirl, pixelate, and/or mosaic effects applied.

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -432,7 +432,12 @@ class Drawable {
         }
 
         // If the effect bits for mosaic, pixelate, whirl, or fisheye are set, use linear
-        if (this._effectBits & 0b0011110) {
+        if (this._effectBits & (
+            ShaderManager.EFFECT_INFO.fisheye.mask |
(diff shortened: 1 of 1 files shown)


src/Drawable.js
```diff
@@ -433,6 +433,16 @@ class Drawable {
             return true;
         }
 
+        // If the effect bits for mosaic, pixelate, whirl, or fisheye are set, use linear
+        if (this._effectBits & (
```
fsih: I think you want parens around the statement before !== 0, because !== is taking precedence over &

adroitwhiz: Good catch

Issue #365, opened 2018-10-29
Scratch spends a fair bit of time in getFencedPosition, can it be faster?


mzgoddard:
Scratch projects change the positions of their sprites very regularly. Since Scratch keeps Sprites within the visible bounds of the rendered stage, getFencedPositionOfDrawable is called for every position change.

There are a few things that could be changed to possibly speed up getFencedPositionOfDrawable.

- getFencedPositionOfDrawable in the convex hull points path creates approximately 8 objects (assuming the hull has 3 points). Most or all of these could use preexisting or cached objects reduce object creation and garbage collection.
- getFencedPositionOfDrawable in the aabb path creates 10 objects.
- getFastBounds could just ensure u_modelMatrix is up to date instead of both it and _inverseMatrix. (In fact both getBounds and getAABB already make sure that u_modelMatrix is up to date.)
- getFastBounds could be assigned to getBounds or getAabb by setConvexHullDirty and setConvexHullPoints

PR #445, opened 2019-04-23
Cache and don't re-create transformed hull points

adroitwhiz:
### Resolves

~~One~~ Two of the concerns in #365 

### Proposed Changes

- Create `Rectangle`s per-`Drawable` for holding `get(Fast)Bounds` results
- Store transformed convex hull points in an array initialized when un-transformed convex hull points are initialized
- Add "transformed convex hull dirty" flag

### Reason for Changes

- Create `Rectangle`s per-`Drawable` for holding `get(Fast)Bounds` results

`Drawable.getBounds`, `Drawable.getBoundsForBubble`, `Drawable.getFastBounds`, and `Drawable.getAABB` take an optional `result` parameter (a `Rectangle`). If given, it will be set to the drawable's bounds.

If a `result` parameter was not passed, however, those functions would create an entirely new `Rectangle` on each call. To make things worse, they're called from several places in the codebase without a `result` being passed.

Now, each `Drawable` instantiates its own `Rectangle`s to hold the bounds results only once: when the `Drawable` itself is created. This should prevent unnecessary creation of `Rectangle`s.

- Store transformed convex hull points in an array initialized when un-transformed convex hull points are initialized

Previously, calling `_getTransformedHullPoints` would result in an entirely new array being created and filled up with `twgl.v3` objects each time.

Now, the array is created and filled with `twgl.v3` objects only when the convex hull itself changes. `_getTransformedHullPoints` now modifies these `twgl.v3` objects in-place. This change should greatly reduce the number of unnecessary objects created.

- Add "transformed convex hull dirty" flag

Previously, calling `_getTransformedHullPoints` meant always recalculating the positions of transformed convex hull points. Now, they are only recalculated when the transform has actually changed.

These changes greatly reduce the allocations made by convex-hull functions:

["Vector scaling performance test 2"](https://scratch.mit.edu/projects/298553174/) (makes heavy use of "if on edge, bounce")

Before | After
-|-
image | image
image | image

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -102,6 +102,12 @@ class Drawable {
         this._convexHullPoints = null;
         this._convexHullDirty = true;
 
+        // The precise bounding box will be from the transformed convex hull points,
+        // so initialize the array of transformed hull points in setConvexHullPoints.
+        // Initializing it once per convex hull recalculation avoids unnecessary creation of twgl.v3 objects.
+        this._transformedHullPoints = null;
+        this._transformedHullDirty = true;
+
         this._skinWasAltered = this._skinWasAltered.bind(this);
     }
 
@@ -120,6 +126,7 @@ class Drawable {
     setTransformDirty () {
         this._transformDirty = true;
         this._inverseTransformDirty = true;
+        this._transformedHullDirty = true;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


adroitwhiz:
This depends on #495

src/Drawable.js
```diff
@@ -591,23 +606,27 @@ class Drawable {
...
+        if (!this._transformedHullDirty) {
+            return this._transformedHullPoints;
+        }
+
         const projection = twgl.m4.ortho(-1, 1, -1, 1, -1, 1);
         const skinSize = this.skin.size;
         const halfXPixel = 1 / skinSize[0] / 2;
         const halfYPixel = 1 / skinSize[1] / 2;
         const tm = twgl.m4.multiply(this._uniforms.u_modelMatrix, projection);
-        const transformedHullPoints = [];
```
cwillisf, line 599: Have you verified that callers never modify the contents of the array returned by this function?

adroitwhiz: `_getTransformedHullPoints` is called by `getBounds` and `getBoundsForBubble`.

`getBounds` passes it to `Rectangle.initFromPointsAABB` which does not modify the array.

`getBoundsForBubble` calls `map` and `filter` on the array, both of which create new arrays instead of modifying in-place.

cwillisf (approved):
👍

PR #448, opened 2019-04-25
Remove auto-generated documentation

### LINKED ISSUE #447: JSDoc is basically unmaintained, and it's started breaking the build
Currently, the ```docs``` npm script uses [jsdoc](https://www.npmjs.com/package/jsdoc), whose last change was pushed to npm **2 years ago**. Now that Travis-CI has been updated to support Node 12, an internal Node function, used by a dependency of JSDoc last updated **5 years ago**, no longer exists, and ```npm run docs``` now fails.

While this particular issue is likely workaround-able, a better long-term solution would be to move to a documentation package which is currently maintained, such as [documentation.js](https://www.npmjs.com/package/documentation).

adroitwhiz:
### Resolves

Resolves #447 

### Proposed Changes

This PR removes the `jsdoc` automatic documentation generator package, and the npm `docs` script.

### Reason for Changes

The quality of the automatically-generated documentation was so bad that it was unusable, and `jsdoc` is no longer maintained anyway. This lack of maintenance is now causing the Travis-CI build to break.

I've looked around for other auto-documentation packages, but all of them somehow managed to output something less readable than what `jsdoc` did. 

I did a quick search and it looks like the only other Scratch repository with auto-generated documentation is scratch-vm, so I would hope it isn't terribly important to have, especially considering its lack of quality.


thisandagain:
Please file an issue for discussion before creating a PR.

PR #451, opened 2019-05-01
Implement canvas-based TextBubbleSkin

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/2721 (I think)

### Proposed Changes

This PR replaces the SVG text bubble generator with a canvas-based `TextBubbleSkin`. This skin is rendered in screen-space and its texture updates synchronously (to the best of my knowledge).

### Reason for Changes

This removes all the SVG rendering quirks (textures only re-rendering at power-of-two scale factors, weird jaggies, etc.) from text bubbles.

It also improves performance; see [Test Bubbles](https://scratch.mit.edu/projects/305895465/). It's important to note that the old implementation will still be faster in full-screen mode, but only due to a bug in the SVG renderer which renders the text bubbles at 1x scale for the first frame they are shown.
However, the new implementation is significantly faster at the same scale (~2x according to my tests, but more benchmarking would be a good idea).

### Test Coverage

I'm not quite sure how to test this! Any guidance is appreciated.


paulkaplan:
Wow this looks great! I am also seeing about a 2x speedup! I tested other language/character sets, line breaks, and it seems to also fix some unusual "jumping" that happens when quickly refreshing the text bubble.

paulkaplan:
@cwillisf not sure if you want to take a peek at this

paulkaplan:
Incidentally, this seems to fix garbled text ordering with multi-line say bubbles: https://github.com/LLK/scratch-vm/issues/1887

PR #461, opened 2019-05-22
Move from UglifyJS to Terser

adroitwhiz:
### Resolves

Closes #345

### Proposed Changes

This replaces `uglifyjs-webpack-plugin` with `terser-webpack-plugin`.

### Reason for Changes

Last September, `uglifyjs-webpack-plugin` switched from `uglify-es`, which supports ES6, to `uglify-js`, which does not. As a result, this codebase won't build with newer versions of `uglifyjs-webpack-plugin`. This is somewhat okay because it's pinned to the old version, but it's a good idea to switch to `terser-webpack-plugin`, which currently supports ES6, in order to stay up-to-date.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,8 +39,8 @@
     "json": "^9.0.4",
     "scratch-vm": "0.2.0-prerelease.20190207224121",
     "tap": "^11.0.0",
+    "terser-webpack-plugin": "^1.2.4",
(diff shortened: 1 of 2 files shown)


fsih:
From Max: Maybe we should get rid of uglify since GUI now builds render

PR #470, opened 2019-06-12
Skin alter push

mzgoddard:
### Resolves

Make switching costumes for lots of clones faster.

### Proposed Changes

- Remove skin altered listeners from Drawable.
- Add skin altered listener to RenderWebGL.
- Push skin alterations down to Drawables currently set to a skin from RenderWebGL.
- Replace skin size getter with a property.
- Cache viewOffset in SVGSkin.
- Check rotation center given to SVGSkin before calling parent.

### Reason for Changes

> - Remove skin altered listeners from Drawable.

This is very expensive for Clones sharing a skin with many other skins. The Drawable's listener has to be found in the list of listeners to remove it.

> - Add skin altered listener to RenderWebGL.

Adding a listener is cheaper than removing, but adding means we need to later remove.

> - Push skin alterations down to Drawables currently set to a skin from RenderWebGL.

Changes made by the GUI cause skin alterations. So skin alterations should not happen when a project is fullscreen or player only. And in the case they do happen, they happen at the rate the user can make changes instead of the rate the vm can instruct the costume to switch.

Given that "event design" that will support both these uses, fast switching, and user alterations, might be to push those alterations down to the Drawable from the RenderWebGL instance.

> - Replace skin size getter with a property.

`size` is used a lot. One use is determining if the rotation center changed in Skin.setRotationCenter. Size only changes when setBitmap, setSVG, or other Skin altering events occur. We can cache the value to decrease time spent creating the size array and later collecting its garbage.

> - Cache viewOffset in SVGSkin.

`viewOffset` is a dynamically created object in SVGRenderer. We use it enough and it only changes when the SVG changes that caching it will help in setRotationCenter performance.

> - Check rotation center given to SVGSkin before calling parent.

SVGSkin.setRotationCenter calls its parent Skin.setRotationCenter. This parent prototype lookup is ending up as a expensive cost to update the rotation center. This adds a check on the value for SVGSkin to avoid that most of the time.

### Benchmark Data

Here is a percent total time table from chrome's profiler, measuring total time over the total time of changeX in 14844969. 14844969 heavily switches costumes for a large number of clones.

| 14844969 switchCostume / changeX | before | after |
| --- | ----: | ----: |
|  scratch3_motion.changeX | 100.00% | 100.00% |
|  scratch3_looks.switchCostume | 25.14% | 9.73% |
|  scratch3_looks._setCostume | 22.92% | 7.86% |
|  rendered-target.setCostume | 21.31% | 7.64% |
|  RenderWebGL.updateDrawableProperties | 17.08% | 6.47% |
|  Drawable.updateProperties | 4.40% | 4.59% |
|  SVGSkin.setRotationCenter | 2.91% | 0.00% |
|  Skin.setRotationCenter | 0.49% | 0.00% |
|  svg-renderer.get viewOffset | 0.15% | 0.00% |
|  Drawable.set skin | 6.76% | 0.00% |
|  events.removeListener | 5.34% | 0.00% |
|  events.addListener | 0.75% | 0.00% |

The change in updateDrawableProperties is what we want to see with this change. The remaining time spent switching costumes is function overhead and improved by the updateDrawable* PR.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -3,6 +3,7 @@ const EventEmitter = require('events');
 const hull = require('hull.js');
 const twgl = require('twgl.js');
 
+const Skin = require('./Skin');
 const BitmapSkin = require('./BitmapSkin');
 const Drawable = require('./Drawable');
 const Rectangle = require('./Rectangle');
@@ -291,6 +292,20 @@ class RenderWebGL extends EventEmitter {
     }
 
     /**
+     * Notify Drawables whose skin is the skin that changed.
+     * @param {Skin} skin - the skin that changed.
+     * @private
+     */
+    _skinWasAltered (skin) {
+        for (let i = 0; i < this._allDrawables.length; i++) {
+            const drawable = this._allDrawables[i];
+            if (drawable && drawable._skin === skin) {
+                drawable._skinWasAltered();
+            }
+        }
+    }
+
+    /**
      * Create a new bitmap skin from a snapshot of the provided bitmap data.
      * @param {ImageData|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement} bitmapData - new contents for this skin.
      * @param {!int} [costumeResolution=1] - The resolution to use for this bitmap.
@@ -302,6 +317,7 @@ class RenderWebGL extends EventEmitter {
         const skinId = this._nextSkinId++;
         const newSkin = new BitmapSkin(skinId, this);
         newSkin.setBitmap(bitmapData, costumeResolution, rotationCenter);
+        newSkin.addListener(Skin.Events.WasAltered, this._skinWasAltered.bind(this, newSkin));
         this._allSkins[skinId] = newSkin;
         return skinId;
     }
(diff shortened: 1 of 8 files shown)


src/RenderWebGL.js
```diff
@@ -290,6 +291,20 @@ class RenderWebGL extends EventEmitter {
         this.emit(RenderConstants.Events.NativeSizeChanged, {newSize: this._nativeSize});
     }
 
+    /**
+     * Notify Drawables whose skin is the skin that changed.
```
cwillisf: I'd suggest adding a comment like this to avoid the possibility that someone might "optimize" this code and undo the performance improvement from this PR :)
```suggestion
     * Notify Drawables whose skin is the skin that changed. This loop may look slow but this method is expected to be
     * called rarely: only on user-driven edits, for example. A previous approach had Drawable registering for Skin's
     * alteration events but it turned out that registering and unregistering for those events took significant time.
```

BryceLTaylor:
Test plan:

Description: This pr updates how switching costumes updates the image on the stage for a sprite or backdrop.  It should be making that process faster.

To test: 
Make a sprite with many costumes
Clone the sprite a lot
Have the sprite and its clones change their costumes
edit the costumes in the paint editor
Edit the costumes so they are bigger
edit the costumes so they are smaller
Add effects to the sprite/clones (color, whirl, etc.)
Change the size of the sprite (bigger, smaller)
Rotate the sprite
Move the rotation center of the sprite (in the paint editor)
Switch the costume from bitmap to vector

Make sure the sprite and its clones look as expected on the stage.
Run the project that changes these properties on the project page
Run it in full screen mode
Run in embed mode

Sometimes doing some of these steps in a different order had caused bugs to be seen or not be seen.  Try doing the tests in different orders.

PR #475, opened 2019-07-08
Set PenSkin silhouette data directly

adroitwhiz:
### Proposed Changes

- Pre-allocate `ImageData` and an array for `gl.readPixels` for `PenSkin`s.
- Set `PenSkin`'s `Silhouette` directly from the skin's `ImageData`.

### Reason for Changes

The previous texture > silhouette pipeline for `PenSkin` was:
1. `gl.readPixels` of the `PenSkin`'s framebuffer into a newly created `Uint8Array`.
2. Create a new `ImageData` object from the array.
3. Draw the `ImageData` to the `PenSkin`'s canvas.
4. Pass the canvas to the `Silhouette`, which:
5. Draws the passed canvas to its own canvas, then
6. Reads its canvas' `ImageData` and stores it.

The new pipeline is:
1. `gl.readPixels`  of the `PenSkin`'s framebuffer into a previously allocated `Uint8Array`.
2. Read the array into a previously created `ImageData` object.
3. Pass the `ImageData` to the `Silhouette`, which uses it directly.

By only creating a new array and `ImageData` object when the `PenSkin`'s canvas size is changed, and passing the `Silhouette` the `ImageData` directly, we can greatly improve the speed of silhouette updates as shown in [this benchmark](https://scratch.mit.edu/projects/320184338/):

| Latest `develop` | This PR |
|-|-|
|image|image|

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -109,6 +109,12 @@ class PenSkin extends Skin {
         /** @type {boolean} */
         this._silhouetteDirty = false;
 
+        /** @type {Uint8Array} */
+        this._silhouettePixels = null;
+
+        /** @type {ImageData} */
+        this._silhouetteImageData = null;
+
         /** @type {object} */
         this._lineOnBufferDrawRegionId = {
             enter: () => this._enterDrawLineOnBuffer(),
@@ -597,6 +603,9 @@ class PenSkin extends Skin {
         gl.clearColor(0, 0, 0, 0);
         gl.clear(gl.COLOR_BUFFER_BIT);
 
+        this._silhouettePixels = new Uint8Array(Math.floor(width * height * 4));
+        this._silhouetteImageData = this._canvas.getContext('2d').createImageData(width, height);
+
         this._silhouetteDirty = true;
     }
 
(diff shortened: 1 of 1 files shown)


src/PenSkin.js
```diff
@@ -649,24 +658,17 @@ class PenSkin extends Skin {
...
-
-            this._silhouette.update(this._canvas);
+            gl.readPixels(
+                0, 0,
+                this._canvasSize[0], this._canvasSize[1],
+                gl.RGBA, gl.UNSIGNED_BYTE, this._silhouettePixels
+            );
+
+            this._silhouetteImageData.data.set(this._silhouettePixels);
+            this._silhouette.update(this._silhouetteImageData);
```
mzgoddard: Its great to see this updated to use the ImageData api change.

src/PenSkin.js
```diff
@@ -597,6 +603,9 @@ class PenSkin extends Skin {
         gl.clearColor(0, 0, 0, 0);
         gl.clear(gl.COLOR_BUFFER_BIT);
 
+        this._silhouettePixels = new Uint8Array(Math.floor(width * height * 4));
```
cwillisf: Nit: the `math.floor` isn't necessary here

adroitwhiz: It seems to be used in every other place in the codebase that constructs a `Uint8Array`, [specifically to prevent issues with Edge](https://github.com/LLK/scratch-render/pull/132).

If I removed it here, I don't think it would cause any trouble (`width` and `height` *should* be integers), and newer versions of Edge (I tested it on Microsoft Edge 44.18362.449.0, Microsoft EdgeHTML 18.18363) have no trouble with floating-point values in typed array constructors, but it'd break the convention that the rest of the codebase is using. Is that okay?

(P.S. it may be worth revisiting this in the future by either testing whether Edge 15, the minimum version [Scratch officially supports](https://scratch.mit.edu/info/faq), supports floating-point values in `TypedArray` constructors, or by bumping the minimum supported Edge version in the FAQ to something newer.)

cwillisf: Oops - thanks for the reminder! I had completely forgotten about that.

PR #478, opened 2019-07-09
Always use getAABB for fence bounds

### LINKED ISSUE #446: Should Skin.getFenceBounds ever call getFastBounds?
### Expected Behavior

```Skin.getFenceBounds()``` is used to determine how far a drawable can move off-stage, and so it should return a consistent value.

### Actual Behavior

```Skin.getFenceBounds()``` calls ```drawable.getFastBounds()```, which returns different values depending on whether the drawable's convex hull exists. While this behavior is overridden in ```BitmapSkin```, ```SVGSkin``` retains it.

I'm not sure whether this could ever become a practical problem. Is it supposed to happen/okay?

### LINKED ISSUE #477: Clicking on a vector sprite in the sprite pane changes how far offscreen it can move
### Expected Behavior

Clicking on a sprite in the sprite pane should not affect its fencing behavior.

### Actual Behavior

When you click on a sprite in the sprite pane, its precise bounds are calculated from its convex hull as part of the blue selection box effect, and for vector sprites, the fence bounds switch over from the AABB bounds to the precise ones. Scratch 2.0 always uses the AABB bounds, [even for vector sprites](https://s2online.github.io/#110878444).

### Steps to Reproduce

In [this project](https://scratch.mit.edu/projects/320308664/):
1. Press space to move the sprite as far to the right as it will go
2. Click "Sprite1" in the sprite pane, so that the blue selection box flashes above it
3. Press space again and notice how the position of the sprite changes

adroitwhiz:
### Resolves

Resolves #477 and resolves #446 

### Proposed Changes

This PR changes the `getFenceBounds` function to use `getAABB` for all types of skins, as opposed to just bitmap skins.

### Reason for Changes

`getFastBounds` does different things depending on external state, leading to inconsistent behavior.
It is also incompatible with Scratch 2.0, [which fences all sprites, including vector ones, based on their AABB](https://s2online.github.io/#110878444).

### Test Coverage

I've added a fencing test which ensures fencing is the same before and after "if on edge, bounce" (which calls `getBounds` internally) is called.


m93a:
Ping! What is the current status of this PR?

fsih:
Testing:
- Make sure to test games where you switch levels/die by walking off the screen
- Clicking on a sprite in a sprite pane, using the if on edge bounce block, used to cause the bounding box of the sprite to shrink by ~half a pixel, but now it should NOT change.

Issue #432, opened 2019-04-11
Reverted Touching Color change because it introduced an issue on Edge


BryceLTaylor:
While testing https://github.com/LLK/scratch-gui/pull/4732 during our bug hunt, @ericrosenbaum noticed that there was an issue introduced on Edge (not on mac chrome) that caused `touching color` to not be true when it was expected to be true.

He created a project (https://scratch.mit.edu/projects/301945007/editor) that exhibited the bug.

His original report says: "On Edge, touching color misses a touch (same project works correctly on staging mac chrome, and works correctly on production on Edge and mac chrome). Test project is here (unshared): "

BryceLTaylor:
Upon further investigation of that project (301945007), when I loaded it on Chrome on Mac the cat didn't say anything while passing the bread.  Then I put the project in fullscreen mode it DID say "Hello" while touching the bread with both feet.  Putting it back into regular mode the cat continues to say "Hello" as expected.  This is on production.

PR #479, opened 2019-07-09
Push out touching-sprite candidate bounds to integer coords

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/2076

### Proposed Changes

This changes `_candidatesTouching`, used for "touching sprite" queries, to expand the bounding boxes of all candidate drawables to integer coordinates, as is already done for the query drawable in `_touchingBounds`.

### Reason for Changes

If a Drawable's bounds lie at a half-pixel, that pixel will still render as partially opaque and should be considered "inside" the drawable. If not, you could end up with e.g. subtle subpixel differences between a sprite's "fast" bounds and its convex hull bounds which result in inconsistent collision detection, as in https://github.com/LLK/scratch-vm/issues/2076.

### Test Coverage

I have added a test for the issue described in https://github.com/LLK/scratch-vm/issues/2076.


src/Silhouette.js
```diff
@@ -140,8 +140,8 @@ class Silhouette {
      * @returns {Uint8ClampedArray} dst
      */
     colorAtLinear (vec, dst) {
```
adroitwhiz: X and Y are still floored here, but that should be okay because we're near a pixel boundary, it'll just interpolate to the next pixel over anyway.

adroitwhiz:
@gnarf since you wrote the CPU renderer code, you may find this of interest.

adroitwhiz:
@fsih https://github.com/LLK/scratch-render/pull/479/commits/b3bcbe4eb9b8aed53553bc75bc231b58c6742ac3 fixes an issue introduced with #515--it was revealed by running tests in both CPU and GPU mode.

I haven't seen any projects "in the wild" that are broken, but if you want me to `git cherry-pick` that commit into a new PR, I'd be happy to do so.

src/RenderWebGL.js
```diff
@@ -1821,12 +1839,12 @@ class RenderWebGL extends EventEmitter {
         // *Not* Scratch Space-- +y is bottom
         // Loop over all rows of pixels, starting at the top
         for (let y = 0; y < height; y++) {
```
fsih, line 1890: Add comment here

fsih:
FYI I tried reproing https://github.com/LLK/scratch-render/issues/432 using https://scratch.mit.edu/projects/301945007/editor on various browsers (but not a surface) and didn't get repro with this patched

fsih:
@cwillisf saw the issue where the cat does not touch the bread on a lower pixel density device, but it was also on prod. We think that the pr that was rolled back was not the cause of this issue, rather it's that the higher density mip map is overwriting the silhouette on higher density devices, and the lower density silhouette doesn't touch the bread. New issue filed: https://github.com/LLK/scratch-render/issues/656

adroitwhiz:
@fsih @cwillisf Any chance of this PR going in? Not sure if you still need to figure out QA...

There are other "touching" fixes I want to make that would be kinda pointless without this one going in first.

adroitwhiz:
Rebased this on the latest develop branch; the only merge conflict I had to resolve was the `_makeOrthoProjection` change in `extractDrawable`, which I moved to `extractDrawableScreenSpace`.

fsih:
Hey @adroitwhiz, really sorry for the lag on this ): unfortunately we don't have the bandwidth right now to put in these changes, so they likely won't be going in soon.

adroitwhiz:
@cwillisf Is there any path to getting this change put in at any point and mitigating any potential project breakage?

Over the years I've had various thoughts about writing my own Scratch runtime or something similar, and every time, I end up going "ugh, but I'll have to replicate 3.0's busted sprite collision code..."

If Scratch itself ever decides to rewrite the renderer (perhaps when rewriting the rest of the editor?), it might have to replicate these bugs too.

I feel like the longer this PR stays in limbo, the more projects will be created that rely on the broken behavior (and it is broken in many, many ways. The "touching" results don't visually match the sprite, the hitbox changes when you click the sprite and its bounds become tighter, and the magnitude of the error grows the smaller the sprite is).

Most projects *should* behave similarly if these fixes go in, but I can think of a couple cases where there may be problems:
- Hairline collisions where a sprite is moving parallel to another sprite's hitbox rather than moving towards or away from it. I don't think (m)any projects do this, but I could be wrong.
- ~~Fencing bugs--some projects may depend on sprites being able to go a certain distance off-screen. This PR should *shrink* hitboxes for the most part, allowing sprites to go further off-screen, but there may be some cases where it doesn't?~~
  I forgot that fencing uses sprites' axis-aligned bounds, which aren't affected by this change. "if on edge, bounce" *does* use convex hull bounds, and its behavior would be slightly affected.

I understand that testing this PR across various projects to see the breakage requires a lot of QA work, but *not* putting this change in also has a cost: you're essentially committing to replicating the broken behavior in all future versions of Scratch.

PR #481, opened 2019-07-12
Return early from EffectTransform methods if no effect bits are set

adroitwhiz:
### Proposed Changes

This PR changes `EffectTransform`'s `transformColor` and `transformPoint` methods to check if any effect bits are set, and if not, return their inputs before doing anything else (e.g. getting uniforms, copying the input to the output).

### Reason for Changes

These functions are called inside tight loops, and optimizing them in this way appears to yield a ~25-30% performance improvement on "touching color" checks, according to [this benchmark](https://scratch.mit.edu/projects/320660090/):

| Current | This PR |
|-|-|
|image|image

### DIFF
--- a/src/EffectTransform.js
+++ b/src/EffectTransform.js
@@ -123,6 +123,10 @@ class EffectTransform {
      * @returns {Uint8ClampedArray} dst filled with the transformed color
      */
     static transformColor (drawable, color4b, dst, effectMask) {
+        const effects = drawable.getEnabledEffects() & effectMask;
+        // If no effects are enabled, don't get anything, create anything, or set anything; just give back the input.
+        if (effects === 0) return color4b;
+
         dst = dst || new Uint8ClampedArray(4);
         effectMask = effectMask || 0xffffffff;
         dst.set(color4b);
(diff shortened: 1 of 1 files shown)


src/EffectTransform.js
```diff
@@ -123,6 +123,10 @@ class EffectTransform {
      * @returns {Uint8ClampedArray} dst filled with the transformed color
      */
     static transformColor (drawable, color4b, dst, effectMask) {
+        const effects = drawable.getEnabledEffects() & effectMask;
+        // If no effects are enabled, don't get anything, create anything, or set anything; just give back the input.
+        if (effects === 0) return color4b;
```
mzgoddard: I like this early return. This will definitely help by skipping the other if statements that would in this case all return false and perform no further work.

However we must move this early exit to after `dst` is initialized and return `dst`. That will match the documentation and the current behaviour.

src/EffectTransform.js
```diff
@@ -187,11 +190,13 @@ class EffectTransform {
      * @return {twgl.v3} dst - The coordinate after being transform by effects.
      */
     static transformPoint (drawable, vec, dst = twgl.v3.create()) {
+        const effects = drawable.getEnabledEffects();
+        // If no effects are enabled, don't get anything, create anything, or set anything; just give back the input.
+        if (effects === 0) return vec;
```
mzgoddard: Same here. Keep the early exit but move it to after `dst` is initialized and return `dst`.

adroitwhiz:
@mzgoddard Good catch!

adroitwhiz:
Closing this PR in favor of a better one

PR #485, opened 2019-07-18
Optimize effect transforms

adroitwhiz:
### Proposed Changes

- Replace `Drawable.getEnabledEffects()` function with `Drawable.enabledEffects`
- Only apply effect transform in `Drawable.getLocalPosition` and `Drawable.sampleColor4b` if drawable effects are enabled
- Replace `possiblyDefinedArgument || defaultValue` pattern with `if (typeof possiblyDefinedArgument === 'undefined') possiblyDefinedArgument = defaultValue` in `EffectTransform` methods

### Reason for Changes

- Replace `Drawable.getEnabledEffects()` function with `Drawable.enabledEffects`
  - The `EffectTransform.transform[Color/Point]` methods, called once per pixel in touching tests, get a Drawable's enabled effects. Replacing the pseudo-getter function with a property speeds it up.
- Only apply effect transform in `Drawable.getLocalPosition` and `Drawable.sampleColor4b` if drawable effects are enabled
  - This is what I did in #481, but taken one step back to the functions that *call* the effect transform methods in question. This alone provides a bigger performance boost than #481.
- Replace `possiblyDefinedArgument || defaultValue` pattern with `if (typeof possiblyDefinedArgument === 'undefined') possiblyDefinedArgument = defaultValue` in `EffectTransform` methods
  - This provides a small performance boost, possibly by removing the "truthy/falsy cast".

This gives a ~40% performance improvement to "touching color" checks on drawables with no effects enabled, and a ~15% performance improvement to ones with effects enabled, as seen in [this benchmark](https://scratch.mit.edu/projects/321498300/):

Current | This PR
-|-
image | image

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -695,7 +695,7 @@ class Drawable {
         // : drawable.skin._silhouette.colorAtLinear(localPosition, dst);
 
         if (drawable.enabledEffects === 0) return textColor;
-        return EffectTransform.transformColor(drawable, textColor, textColor);
+        return EffectTransform.transformColor(drawable, textColor);
     }
 }
 
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -36,8 +36,12 @@ const getLocalPosition = (drawable, vec) => {
     // localPosition matches that transformation.
     localPosition[0] = 0.5 - (((v0 * m[0]) + (v1 * m[4]) + m[12]) / d);
     localPosition[1] = (((v0 * m[1]) + (v1 * m[5]) + m[13]) / d) + 0.5;
-    // Apply texture effect transform if the localPosition is within the drawable's space.
-    if ((localPosition[0] >= 0 && localPosition[0] < 1) && (localPosition[1] >= 0 && localPosition[1] < 1)) {
+    // Apply texture effect transform if the localPosition is within the drawable's space,
+    // and any effects are currently active.
+    if (drawable.enabledEffects !== 0 &&
+        (localPosition[0] >= 0 && localPosition[0] < 1) &&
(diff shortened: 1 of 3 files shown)


src/Drawable.js
```diff
@@ -35,8 +35,12 @@ const getLocalPosition = (drawable, vec) => {
     // localPosition matches that transformation.
     localPosition[0] = 0.5 - (((v0 * m[0]) + (v1 * m[4]) + m[12]) / d);
     localPosition[1] = (((v0 * m[1]) + (v1 * m[5]) + m[13]) / d) + 0.5;
-    // Apply texture effect transform if the localPosition is within the drawable's space.
-    if ((localPosition[0] >= 0 && localPosition[0] < 1) && (localPosition[1] >= 0 && localPosition[1] < 1)) {
+    // Apply texture effect transform if the localPosition is within the drawable's space,
+    // and any effects are currently active.
+    if (drawable.enabledEffects !== 0 &&
```
adroitwhiz: this will conflict with https://github.com/LLK/scratch-render/pull/479/files#diff-3dbe88c4ff1ea0a8c0b27759902040f5R44

src/EffectTransform.js
```diff
@@ -123,15 +123,17 @@ class EffectTransform {
      * @returns {Uint8ClampedArray} dst filled with the transformed color
      */
     static transformColor (drawable, color4b, dst, effectMask) {
```
cwillisf: Since we only call this from one place, what do you think about simplifying even further? Maybe something like this:
```suggestion
     static transformColor (drawable, inOutColor) {
```

adroitwhiz: Should I remove the `effectMask` parameter too?

PR #487, opened 2019-07-25
penStamp() directly to the PenSkin's framebuffer

adroitwhiz:
### Proposed Changes

This PR changes `RenderWebGL.penStamp` to draw directly to the `PenSkin`'s framebuffer, instead of drawing to the query buffer then drawing the query buffer to the `PenSkin`'s framebuffer as it currently does.

### Reason for Changes

This improves pen stamp performance by ~2.5-3x, according to [this benchmark](https://scratch.mit.edu/projects/322227433/):

Current | This PR
-|-
image | image


griffpatch:
Oh man... How I want stamping to be more efficient! +1 for this kind of optimisation!

griffpatch:
Got to say, of this change is at all possible then I would say it is a wonderful win... Being able to stamp faster would be a just amazing. Optimising the change costume, stamp, change costume, stamp loops to be as fast as possible world improve a lot of projects.

PR #489, opened 2019-08-02
Fix "touching color" for background color

### LINKED ISSUE #488: "Touching color (white)" behaves inconsistently
### Expected Behavior

"touching color (white)" should always return `true` over a white backdrop.

### Actual Behavior

On the CPU path, "touching color (white)" only returns true when the sprite calling it is within the bounds of another drawable. On the GPU path, it never returns true.

### Steps to Reproduce

See [this project](https://scratch.mit.edu/projects/322985073/).

adroitwhiz:
### Resolves

Resolves #488 

### Proposed Changes

This PR makes several changes that are all necessary to properly fix "touching color" for background colors:

- `RenderWebGL._backgroundColor` has been replaced with `RenderWebGL._backgroundColor4f` and `RenderWebGL._backgroundColor3b`. This makes it easier to call `colorMatches` on the background color.
- `RenderWebGL.isTouchingColor` now checks whether the color to be checked `colorMatches` the background color, and if it does, will not limit the check to "candidate" drawables' bounds, as the background spans the entire stage and should always be checked.
- A new `background` draw mode and draw region has been added, which draws the background color as a quad. This is necessary to support the next change.
    - As part of adding a new draw mode, I cleaned up the `#ifdef` logic in the shader code a bit.
- The GPU path for `isTouchingColor` now clears to transparent black, and only draws in the opaque background for pixels masked in by the stencil buffer (which can only be done if the background is drawn as a quad--`gl.clear` will fill the entire buffer). The loop in `_isTouchingColorGpuFin` that checks for a pixel match will no longer check transparent pixels. This is the correct behavior, as only masked-in pixels should be checked.
- `_candidatesTouching`, used for "touching sprite" queries, expands the bounding boxes of all candidate drawables to integer coordinates, as is already done for the query drawable in `_touchingBounds`.
  - This was originally only part of #479, but since we're looping over the pixels contained within the bounds returned by `_candidatesBounds`, and `_candidatesBounds` uses the bounding boxes returned by `_candidatesTouching`, it's extra important that the bounds be integers-- other code expects it to be so. For instance, [what do you expect this code to do if `bounds.width` and `bounds.height` are floating-point values?](https://github.com/LLK/scratch-render/blob/77ccf978864082c4fb6c634925609c5a31851891/src/RenderWebGL.js#L870-L871)

### Reason for Changes

These changes are all necessary to make "touching color" behave properly when checking if the background color is being touched.

### Test Coverage

An "is touching background color" test has been added.


src/RenderWebGL.js
```diff
@@ -842,11 +875,16 @@ class RenderWebGL extends EventEmitter {
             };
         }
 
+        // TODO: figure out why this is all in a try block?
```
adroitwhiz: IIRC this try-finally dates back to cwillisf's initial implementation of the code. What error is it supposed to catch?

adroitwhiz:
This is actually failing the test I wrote on the GPU mode. At least we know the test works...

adroitwhiz:
Fixed it! I'm surprised we didn't see issues earlier-- in some places we were initializing loop counters to floating-point values!

src/RenderWebGL.js
```diff
@@ -182,9 +182,23 @@ class RenderWebGL extends EventEmitter {
...
         this._snapshotCallbacks = [];
 
+        /** @type {Array<number>}
+         * @readonly */
+        this._backgroundColor4f = [0, 0, 0, 0];
+
+        /** @type {Uint8ClampedArray}
+         * @readonly */
+        this._backgroundColor3b = new Uint8ClampedArray(3);
+
```
adroitwhiz, line 201: I didn't remove `this._backgroundColor` because it never existed here-- it's first defined when `setBackgroundColor` is called

src/RenderWebGL.js
```diff
@@ -244,7 +258,12 @@ class RenderWebGL extends EventEmitter {
      * @param {number} blue The blue component for the background.
      */
     setBackgroundColor (red, green, blue) {
-        this._backgroundColor = [red, green, blue, 1];
+        this._backgroundColor4f = [red, green, blue, 1];
```
cwillisf: I think this might generate less GC work:
```suggestion
        this._backgroundColor4f[0] = red;
        this._backgroundColor4f[1] = green;
        this._backgroundColor4f[2] = blue;
        this._backgroundColor4f[3] = 1;
```

src/RenderWebGL.js
```diff
@@ -244,7 +258,12 @@ class RenderWebGL extends EventEmitter {
      * @param {number} blue The blue component for the background.
      */
     setBackgroundColor (red, green, blue) {
```
fsih, line 260: Let's consider removing this function entirely in a future PR

src/RenderWebGL.js
```diff
@@ -182,9 +182,23 @@ class RenderWebGL extends EventEmitter {
...
+            enter: () => this._enterDrawBackground(),
+            exit: () => this._exitDrawBackground()
+        };
+
         /** @type {Array.<snapshotCallback>} */
         this._snapshotCallbacks = [];
 
+        /** @type {Array<number>}
+         * @readonly */
+        this._backgroundColor4f = [0, 0, 0, 0];
```
fsih: Because this says readonly, it looks to me like background color is black by default and will always be black, so might be better to have these initialize to white and drop the call to setBackgroundColor

src/RenderWebGL.js
```diff
@@ -1292,6 +1344,13 @@ class RenderWebGL extends EventEmitter {
                     // Update the CPU position data
                     drawable.updateCPURenderAttributes();
                     const candidateBounds = drawable.getFastBounds();
+
+                    // Push bounds out to integers. If a drawable extends out into half a pixel, that half-pixel stil
```
fsih: typo still

PR #496, opened 2019-08-13
Revert "Revert "Skin alter push""

kchadha:
Reverts LLK/scratch-render#493

This PR needs investigation as to whether this original PR (e.g. #470) or #467 was failing gui integration tests.

Fix (if necessary) and re-make this PR.

cc/ @cwillisf @mzgoddard

PR #497, opened 2019-08-13
Set SVG skin size synchronously

adroitwhiz:
### Resolves

Solves `SVGSkin`s' size staying small; subsequently supersedes #496, #486

This PR fixes SVG costumes' size showing as 0x0 in the editor, which was causing the GUI's SVG costume integration tests to fail as they were specifically searching for the string `100 x 100`.

### Proposed Changes

This PR re-merges "Skin alter push", but changes `SVGSkin.setSVG()` to set the costume `size` outside of the `SVGRenderer` callback rather than inside it.

It also includes a minor change (previously #486), which swaps the order of setting the skin `size` and calling `calculateRotationCenter`. This fixes calculated rotation centers.

### Reason for Changes

The VM expects the skin size to be properly set immediately after calling `updateSVGSkin`, as it sets the costume size to the skin size after calling it. If the skin size is only set after the SVG renderer callback fires, the SVG costume size will appear as 0x0 in the editor.

### Test Coverage

This appears to be, in a way, tested for in the GUI. A more explicit test over there might be a good idea.

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -114,7 +114,19 @@ class SVGSkin extends Skin {
      * @fires Skin.event:WasAltered
      */
     setSVG (svgData, rotationCenter) {
-        this._svgRenderer.fromString(svgData, 1, () => {
+        this._svgRenderer.loadString(svgData);
+
+        // Size must be updated synchronously because the VM sets the costume's `size` immediately after calling this.
+        // TODO: add either a callback to this function so the costume size can be set after this is done,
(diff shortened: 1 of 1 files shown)


src/SVGSkin.js
```diff
@@ -100,7 +114,19 @@ class SVGSkin extends Skin {
      * @fires Skin.event:WasAltered
      */
     setSVG (svgData, rotationCenter) {
-        this._svgRenderer.fromString(svgData, 1, () => {
+        this._svgRenderer.loadString(svgData);
```
adroitwhiz: my additional changes start here

Issue #500, opened 2019-08-21
Black and white dots appearing on stage on Chromebook


fsih:
Several users are reporting seeing black dots on stage.
I'm seeing this issue on Samsung Chromebook 3, Chrome version 76. Was NOT seeing it on Chrome version 75.
Not able to repro on Mac 10.14, Chrome 76
Not able to repro on Windows 10, Chrome 76

The issue is visible at all stage scales, whether in the editor or on the project page. The dots are not visible in the paint editor. The dots appear regardless of the color of the costume or whether the costume is bitmap or vector. Dots change location when the sprite is dragged to a different place on the stage.

![Screenshot 2019-08-21 at 11 46 18 AM](https://user-images.githubusercontent.com/2855464/63454862-a2703580-c419-11e9-8be8-26a5429c2183.png)
![Screenshot 2019-08-21 at 11 45 36 AM (1)](https://user-images.githubusercontent.com/2855464/63454864-a2703580-c419-11e9-924d-cd0e417c78e7.png)
![Screenshot 2019-08-21 at 11 45 59 AM](https://user-images.githubusercontent.com/2855464/63454859-a2703580-c419-11e9-9d82-7d7674f945ce.png)
![Screenshot 2019-08-21 at 11 50 48 AM](https://user-images.githubusercontent.com/2855464/63455368-a81a4b00-c41a-11e9-9b48-066416d36706.png)

fsih:
A couple of user threads
https://scratch.mit.edu/discuss/topic/367470/
https://scratch.mit.edu/discuss/topic/367460/

fsih:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=996323 on Chrome

fsih:
Some comments from @paulkaplan 
a couple other observations: it isn’t in the paint editor, meaning it isn’t a 2d canvas problem. It also isn’t in the image you get when you drag the cat around on the stage, which is kinda wild because that is pixels retrieved from the renderer itself. It is present on the thumbnail (which makes sense, that is just the canvas converted to pixels)

fsih:
https://bugs.chromium.org/p/chromium/issues/detail?id=996323#c2 Chrome responded on the bug and wants a minimum repro. @cwillisf any chance there is a very stripped-down version of the renderer that I could start from?

adroitwhiz:
Does the issue appear in the "playground"?

paulkaplan:
@adroitwhiz it does, yeah

[...]

PR #502, opened 2019-08-28
Turn off antialias

fsih:
Fixes https://github.com/LLK/scratch-render/issues/500

Also, antialias isn't supported on all browsers, so this should make behavior more consistent. The downside is that vectors will have pixely edges, especially on lower density devices

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -104,7 +104,7 @@ class RenderWebGL extends EventEmitter {
      * @private
      */
     static _getContext (canvas) {
-        return twgl.getWebGLContext(canvas, {alpha: false, stencil: true});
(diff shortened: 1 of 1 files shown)


adroitwhiz:
To be clear, only the bounding-box edges are pixely:
image

adroitwhiz:
This goes a bit deeper than I thought.

Looking at the [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=996323) which this is intended to work around, it appears that the problem's root cause is a bug in the code for CMAA, a form of morphological antialiasing.

Morphological antialiasing is essentially a post-processing filter run on the final render, and intended for 3D scenes. Because WebGL antialiasing is vendor-specific, certain platforms may be running *their own post-processing filters* on the stage when antialiasing is enabled!

PR #509, opened 2019-09-26
Revert "Revert "Put Skin Alter Push Back In""

kchadha:
Reverts LLK/scratch-render#508

Changes from #470 were taken out again because of a bug where making costume edits in the paint editor makes the costume change size on the stage:

skin-alter-push-2

skin-alter-push

cc/ @mzgoddard, @cwillisf

### DIFF
--- a/src/BitmapSkin.js
+++ b/src/BitmapSkin.js
@@ -23,6 +23,12 @@ class BitmapSkin extends Skin {
 
         /** @type {Array<int>} */
         this._textureSize = [0, 0];
+
+        /**
+         * The "native" size, in texels, of this skin.
+         * @type {Array<number>}
+         */
+        this.size = [0, 0];
     }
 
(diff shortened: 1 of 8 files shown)


mzgoddard:
Ah. So each draw calls updateBitmapSkin or updateSVGSkin. When the call is on a skin of the other type it'll create a new skin of the right type, but I overlooked binding the event listener here. So the first draw in the new type, will update the drawable matrix because the drawable's skin changes, reseting the matrix. When the second draw happens, the drawable draws the right skin, but the matrix isn't updated as it doesn't get a message from the skin changing, making it draw in the small first matrix.

towerofnix:
So, um.

Revert "Revert "Revert "Revert "Revert "Revert "Skin alter push"""""". :package:

I'll stop now [:

PR #515, opened 2019-10-13
Use premultiplied alpha everywhere, take 2

### LINKED ISSUE #237: Shapes have fuzzy black outlines on stage
### Expected Behavior

White shapes without outlines should have no outlines

### Actual Behavior

White shapes have fuzzy black outlines on the stage
image

### Steps to Reproduce

Draw some white things in the paint editor and observe them on the stage.
I think this is happening for non-white shapes too, it's just harder to tell
image

Windows 10 Chrome

### LINKED ISSUE #459: Ugly dark fringe around resampled costume's edges

### LINKED ISSUE #513: Ghost effect does not work on premultiplied alpha skins
See https://github.com/LLK/scratch-vm/pull/2291 for more explanation of why this came up, but basically there is an implicit assumption that was introduced when we introduced the `hasPremultipliedAlpha` flag for skins, that the ghost effect (and possibly other color effects) would not be used on the PenSkin (the only hasPremultipliedAlpha=true skin type). It is implicit because the shader modifies the alpha for ghost in a way that assumes straight alpha composition. There are a number of related issues and proposed solutions, and I don't expect us to have the bandwidth to address them now, but wanted to have a place to track this, given the amount of time invested in understanding https://github.com/LLK/scratch-vm/pull/2291

Related issues/work
- https://github.com/LLK/scratch-render/issues/459
- https://github.com/LLK/scratch-gui/issues/4888
- https://github.com/LLK/scratch-render/pull/460
- https://github.com/LLK/scratch-vm/pull/2291
- https://github.com/LLK/scratch-render/issues/237

adroitwhiz:
### Resolves

Resolves #513 
Resolves #237 

### Proposed Changes

- Set `gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL` before  calling `texImage2D`
  - Move `_texture` to `Skin` base class
  - Add `_setTexture` helper function to `Skin` base class
  - Disable "automatically attempt mipmaps" in `twgl.createTexture`
- Use `gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA)` for all blending
- Premultiply alpha in ghost effect and pen shader
- Remove `stamp` draw mode
- Move software-pipeline alpha premultiplication step from `RenderWebGL.sampleColor3b` to `Silhouette.getColor` methods
  - Allow `Silhouette`s to take both non-premultiplied and premultiplied image data by splitting `getColor4b` into `getColor4b` and `getPremultipliedColor4b`

### Reason for Changes

- Set `gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL` before binding textures

This causes textures' alpha channels to be premultiplied before being sent to the GPU. This is necessary for proper blending around edges of sprites; see [this article ("GPUs prefer premultiplication")](http://www.realtimerendering.com/blog/gpus-prefer-premultiplication/) for more info

- Move `_texture` to `Skin` base class
- Add `_setTexture` helper function to `Skin` base class

These changes reduce code duplication.
`_texture` was previously initialized by every implementation of `Skin`, so I moved it to the base class to reduce redundancy.
`_setTexture` performs the steps of binding the texture, setting and unsetting `gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL`, calling `texImage2D`, and updating the skin's silhouette. This would otherwise have to be done in every place that sets a `Skin` implementation's texture to an image.

- Disable "automatically attempt mipmaps" in `twgl.createTexture`

Because `Skin` textures are non-power-of-two, no mipmaps can be generated, but for some reason twgl will attempt anyway unless explicitly told not to. This change makes `twgl.createTexture` throw out tons of `glGenerateMipmap` warning messages because it is no longer passed a texture (the texture is now set outside the function that creates it).

- Use `gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA)` for all blending

This is the proper blend function for compositing premultiplied textures.

- Premultiply alpha in ghost effect and pen shader

Everything should be premultiplied.

- Remove `stamp` draw mode

The `stamp` draw mode has not done what it's been said to do ("Draw normally except for pre-multiplied alpha") since #418. It has, in fact, done nothing differently than the `default` draw mode.

- Move software-pipeline alpha premultiplication step from `RenderWebGL.sampleColor3b` to `Silhouette.getColor` methods

As described in the article I linked above, color values must be multiplied by alpha *before* interpolation to avoid fringing artifacts.

- Allow `Silhouette`s to take both non-premultiplied and premultiplied image data by splitting `getColor4b` into `getColor4b` and `getPremultipliedColor4b`

A `Silhouette` can take either data from an image, which is not premultiplied, or data from the GPU, which is premultiplied. If the data does not come premultiplied, we want to premultiply it. If it does come premultiplied, we don't want to multiply it again.

I accomplished this by adding the `premultiplied` getter/setter to `Silhouette`. If set to `true`, it indicates that data passed to the silhouette already comes premultiplied, and if set to `false` it indicates that it is not premultiplied.

I added a `_getColor` member to `Silhouette` which the `Silhouette.colorAt` functions now call instead of `getColor4b`. If the silhouette is `premultiplied`, the member will point towards the `getPremultipliedColor4b` function. If not, it will point towards the `getColor4b` function.

This does not appear to regress performance of `Silhouette` operations.

### Unanswered Questions
- It was previously determined that [setting textures from `ImageData` was faster than setting them from `HTMLCanvasElement`s](https://github.com/LLK/scratch-render/pull/414). This may have been because `ImageData` was not premultiplied, and the textures were not either, allowing the browser to copy the texture data to the GPU directly rather than dividing each individual pixel by its alpha.
With this PR, textures are now premultiplied, like an `HTMLCanvasElement`'s backing store. **So is it faster to set premultiplied textures from a canvas instead of grabbing the `ImageData`?**

- ~~The current shader applies the "ghost" effect (which now multiplies all color values) before the "brightness" and "color" effects (which read color values). This appears to produce correct results, but is probably incorrect. However, I cannot find a way to fix it without spaghettifying the code. **How should this be handled? Should the shader divide alpha values if brightness/color are enabled and multiply them again afterwards?**~~
    - Regarding the above, I have decided to divide by alpha values before applying brightness/color effects and multiply afterwards.

- Both in the current code and here, `extractDrawable` returns a premultiplied image. This produces a dark fringe around dragged sprites. **Is this something to fix in the future?**

### DIFF
--- a/src/EffectTransform.js
+++ b/src/EffectTransform.js
@@ -130,15 +130,21 @@ class EffectTransform {
         const effects = drawable.enabledEffects;
         const uniforms = drawable.getUniforms();
 
-        if ((effects & ShaderManager.EFFECT_INFO.ghost.mask) !== 0) {
-            // gl_FragColor.a *= u_ghost
-            inOutColor[3] *= uniforms.u_ghost;
-        }
-
         const enableColor = (effects & ShaderManager.EFFECT_INFO.color.mask) !== 0;
         const enableBrightness = (effects & ShaderManager.EFFECT_INFO.brightness.mask) !== 0;
 
         if (enableColor || enableBrightness) {
+            // gl_FragColor.rgb /= gl_FragColor.a + epsilon;
+            // Here, we're dividing by the (previously pre-multiplied) alpha to ensure HSV is properly calculated
+            // for partially transparent pixels.
+            // epsilon is present in the shader because dividing by 0 (fully transparent pixels) messes up calculations.
+            // We're doing this with a Uint8ClampedArray here, so dividing by 0 just gives 255. We're later multiplying
+            // by 0 again, so it won't affect results.
+            const alpha = inOutColor[3] / 255;
+            inOutColor[0] /= alpha;
+            inOutColor[1] /= alpha;
+            inOutColor[2] /= alpha;
+
             // vec3 hsl = convertRGB2HSL(gl_FragColor.xyz);
             const hsl = rgbToHsl(inOutColor);
 
@@ -171,6 +177,20 @@ class EffectTransform {
             }
             // gl_FragColor.rgb = convertHSL2RGB(hsl);
             inOutColor.set(hslToRgb(hsl));
+
+            // gl_FragColor.rgb *= gl_FragColor.a + epsilon;
+            // Now we're doing the reverse, premultiplying by the alpha once again.
+            inOutColor[0] *= alpha;
+            inOutColor[1] *= alpha;
+            inOutColor[2] *= alpha;
+        }
+
+        if ((effects & ShaderManager.EFFECT_INFO.ghost.mask) !== 0) {
+            // gl_FragColor *= u_ghost
+            inOutColor[0] *= uniforms.u_ghost;
+            inOutColor[1] *= uniforms.u_ghost;
+            inOutColor[2] *= uniforms.u_ghost;
+            inOutColor[3] *= uniforms.u_ghost;
         }
 
         return inOutColor;
--- a/src/Silhouette.js
+++ b/src/Silhouette.js
@@ -39,6 +39,7 @@ const __cornerWork = [
 
 /**
  * Get the color from a given silhouette at an x/y local texture position.
+ * Multiply color values by alpha for proper blending.
  * @param {Silhouette} The silhouette to sample.
  * @param {number} x X position of texture (0-1).
  * @param {number} y Y position of texture (0-1).
@@ -46,6 +47,30 @@ const __cornerWork = [
  * @return {Uint8ClampedArray} The dst vector.
  */
 const getColor4b = ({_width: width, _height: height, _colorData: data}, x, y, dst) => {
+    // 0 if outside bouds, otherwise read from data.
+    if (x >= width || y >= height || x < 0 || y < 0) {
+        return dst.fill(0);
+    }
+    const offset = ((y * width) + x) * 4;
+    // premultiply alpha
+    const alpha = data[offset + 3] / 255;
+    dst[0] = data[offset] * alpha;
+    dst[1] = data[offset + 1] * alpha;
+    dst[2] = data[offset + 2] * alpha;
+    dst[3] = data[offset + 3];
+    return dst;
+};
+
+/**
+ * Get the color from a given silhouette at an x/y local texture position.
+ * Do not multiply color values by alpha, as it has already been done.
+ * @param {Silhouette} The silhouette to sample.
+ * @param {number} x X position of texture (0-1).
+ * @param {number} y Y position of texture (0-1).
+ * @param {Uint8ClampedArray} dst A color 4b space.
+ * @return {Uint8ClampedArray} The dst vector.
+ */
+const getPremultipliedColor4b = ({_width: width, _height: height, _colorData: data}, x, y, dst) => {
     // 0 if outside bouds, otherwise read from data.
     if (x >= width || y >= height || x < 0 || y < 0) {
         return dst.fill(0);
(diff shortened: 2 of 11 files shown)


adroitwhiz:
Pages 20-21 of [WebGL Insights](https://github.com/WebGLInsights/WebGLInsights.github.io/releases/download/v1.0/WebGL.Insights.-.Patrick.Cozzi.pdf) go into a bit more detail on the performance impact of `UNPACK_PREMULTIPLY_ALPHA_WEBGL`.

src/EffectTransform.js
```diff
@@ -133,15 +133,19 @@ class EffectTransform {
...
 
-        if ((effects & ShaderManager.EFFECT_INFO.ghost.mask) !== 0) {
-            // gl_FragColor.a *= u_ghost
-            dst[3] *= uniforms.u_ghost;
-        }
-
         const enableColor = (effects & ShaderManager.EFFECT_INFO.color.mask) !== 0;
         const enableBrightness = (effects & ShaderManager.EFFECT_INFO.brightness.mask) !== 0;
 
         if (enableColor || enableBrightness) {
```
fsih: Can you comment here to say that what we're doing is removing the premultiplied alpha because color and brightness effects need to be applied to non-premultiplied-alpha colors?

src/EffectTransform.js
```diff
@@ -174,6 +178,19 @@ class EffectTransform {
             }
             // gl_FragColor.rgb = convertHSL2RGB(hsl);
             dst.set(hslToRgb(hsl));
+
+            // gl_FragColor.rgb *= gl_FragColor.a + epsilon;
+            dst[0] *= alpha;
```
fsih: Can you add a comment here saying that we're re-applying pre-multiplied alpha?

src/EffectTransform.js
adroitwhiz: :+1:

src/EffectTransform.js
adroitwhiz: :+1:

src/Silhouette.js
```diff
@@ -39,13 +39,38 @@ const __cornerWork = [
...
+ * Get the color from a given silhouette at an x/y local texture position.
+ * Do not multiply color values by alpha, as it has already been done.
+ * @param {Silhouette} The silhouette to sample.
+ * @param {number} x X position of texture (0-1).
+ * @param {number} y Y position of texture (0-1).
+ * @param {Uint8ClampedArray} dst A color 4b space.
+ * @return {Uint8ClampedArray} The dst vector.
+ */
+const getPremultipliedColor4b = ({_width: width, _height: height, _colorData: data}, x, y, dst) => {
     // 0 if outside bouds, otherwise read from data.
```
cwillisf:
```suggestion
     // 0 if outside bounds, otherwise read from data.
```

src/Silhouette.js
```diff
@@ -78,9 +103,45 @@ class Silhouette {
...
+    }
+
+    /**
+     * Set the alpha premultiplication state of this silhouette, to ensure proper color values are returned.
+     * If set to true, the silhouette will assume it is being set with premultiplied color data,
+     * and will not multiply color values by alpha.
+     * If set to false, it will multiply color values by alpha.
+     * @param {boolean} isPremultiplied Whether this silhouette will be populated with premultiplied color data.
+     */
+    set premultiplied (isPremultiplied) {
```
cwillisf: It might make sense to add this as a parameter to the `update()` method since it's a property of the pixel data, not really a property of the silhouette as seen from the "outside" -- the silhouette always tries to return premultiplied color data.

src/shaders/sprite.frag
```diff
@@ -153,16 +153,12 @@ void main()
 
 	gl_FragColor = texture2D(u_skin, texcoord0);
 
-    #ifdef ENABLE_ghost
-    gl_FragColor.a *= u_ghost;
-    #endif // ENABLE_ghost
+	#if defined(ENABLE_color) || defined(ENABLE_brightness)
+	// Divide premultiplied alpha values for proper color processing
+	// Add epsilon to avoid dividing by 0 for fully transparent pixels
+	gl_FragColor.rgb /= gl_FragColor.a + epsilon;
```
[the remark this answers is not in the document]
cwillisf: Also: currently `epsilon` is only defined if the color effect is enabled. @fsih found that if you use the brightness effect without the color effect, this line of the shader doesn't compile. I think it's OK to move the definition of `epsilon` outside of any conditionals.

adroitwhiz:
I think that's all the feedback addressed; thanks!

adroitwhiz:
P.S. any idea how @mzgoddard benchmarks loading speed consistently? I may try reverting #414 per the *WebGL Insights* analysis I linked earlier.

PR #516, opened 2019-10-15
Revert "remove updateMatrix call from getFastBounds"

kchadha:
Reverts LLK/scratch-render#468

The PR caused a regression with video sensing where video motion on sprite reports 0 if the sprite has moved from its original position.


kchadha:
Closed because alternative solution (LLK/scratch-vm#2293) was merged instead.

PR #522, opened 2019-11-01
Remove `SvgRenderer.fromString` call from `SVGSkin.setSVG`

### LINKED ISSUE #518: The call to svg-renderer._draw in SVGSkin.setSVG isn't necessary
During SVGSkin.setSVG, we call
[this._svgRenderer.fromString(svgData, 1, () => {...})](https://github.com/LLK/scratch-render/commit/89b5ed32d06ab8fcd2d83b6dffd396d67afe5118#diff-df3c8bc1ea694143c4638f12818d8611R103)
which triggers a call to svg-renderer._draw at a texture scale of 1 (hard-coded).

Milliseconds later, as a part of vm's runtime._step method, we call draw on all drawables, which also calls svg-renderer._draw, this time at the correct texture scale (one that takes into account the device pixel density.) That call to _draw immediately replaces the previous texture.

Here are the 2 stack traces
Screen Shot 2019-10-30 at 15 55 29

This should mean that the first call to _draw with scale 1 is unnecessary. However, setSVG is only happening whenever a user action happens in the paint editor right now, which is not very often (and so probably not a high priority).

adroitwhiz:
Depends on https://github.com/LLK/scratch-svg-renderer/pull/107

### Resolves

Resolves #518

### Proposed Changes

This PR removes the call to `SVGSkin._svgRenderer.fromString` from the `SVGSkin.loadSVG` method, in the process updating `SVGSkin` to use the new methods added in https://github.com/LLK/scratch-svg-renderer/pull/107.

### Reason for Changes

Removing the `SVGSkin._svgRenderer.fromString` call prevents the skin from being unnecessarily drawn twice (see #518 for more details).

Updating `SVGSkin` to use the new `SvgRenderer` methods makes the code much easier to reason about by removing unpredictable behavior (see [this comment](https://github.com/LLK/scratch-render/issues/518#issuecomment-548943502) and https://github.com/LLK/scratch-svg-renderer/pull/107 for more information).

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -22,6 +22,9 @@ class SVGSkin extends Skin {
         /** @type {SvgRenderer} */
         this._svgRenderer = new SvgRenderer();
 
+        /** @type {boolean} */
+        this._svgDirty = false;
+
         /** @type {WebGLTexture} */
         this._texture = null;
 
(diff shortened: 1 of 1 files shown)


adroitwhiz:
Closing in favor of #527

PR #524, opened 2019-11-09
Reduce unnecessary traversal

Ryuurock:
### Resolves
Resolves https://github.com/LLK/scratch-render/issues/640
Optimized the pick method to traverse the drawable

### Proposed Changes

Optimized the pick method to traverse the drawable

### Reason for Changes

Increase the speed of the pick

### Test Coverage

chrome

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1012,21 +1012,25 @@ class RenderWebGL extends EventEmitter {
      * RenderConstants.ID_NONE if there is no Drawable at that location.
      */
     pick (centerX, centerY, touchWidth, touchHeight, candidateIDs) {
+        const bounds = this.clientSpaceToScratchBounds(centerX, centerY, touchWidth, touchHeight);
+        if (bounds.left === -Infinity || bounds.bottom === -Infinity) {
+            return false;
+        }
+
         candidateIDs = (candidateIDs || this._drawList).filter(id => {
             const drawable = this._allDrawables[id];
             // default pick list ignores visible and ghosted sprites.
             if (drawable.getVisible() && drawable.getUniforms().u_ghost !== 0) {
+                const drawableBounds = drawable.getFastBounds();
+                const inRange = bounds.intersects(drawableBounds);
+                if (!inRange) return false;
+
                 drawable.updateCPURenderAttributes();
                 return true;
             }
             return false;
         });
         if (candidateIDs.length === 0) {
-            return false;
-        }
-
-        const bounds = this.clientSpaceToScratchBounds(centerX, centerY, touchWidth, touchHeight);
-        if (bounds.left === -Infinity || bounds.bottom === -Infinity) {
             return false;
         }
 


src/RenderWebGL.js
```diff
@@ -1002,10 +1002,19 @@ class RenderWebGL extends EventEmitter {
      * RenderConstants.ID_NONE if there is no Drawable at that location.
      */
     pick (centerX, centerY, touchWidth, touchHeight, candidateIDs) {
+        const bounds = this.clientSpaceToScratchBounds(centerX, centerY, touchWidth, touchHeight);
+
         candidateIDs = (candidateIDs || this._drawList).filter(id => {
             const drawable = this._allDrawables[id];
+            const drawableBounds = this.getBounds(id);
```
adroitwhiz: The `getBounds` method updates the convex hull on the target drawable, which is a very slow operation.
Maybe use `const drawableBounds = drawable.getFastBounds();` instead?

src/RenderWebGL.js
```diff
@@ -1002,10 +1002,19 @@ class RenderWebGL extends EventEmitter {
...
+            const inRange = (
+                bounds.top < drawableBounds.top
+                    && bounds.right < drawableBounds.right
+                    && bounds.bottom > drawableBounds.bottom
+                    && bounds.left > drawableBounds.left
+            );
```
adroitwhiz: This check could fail in cases like these:
image
The black line is the top of `drawableBounds`, and the red line is the top of `bounds`.

It looks like the current check only succeeds if `targetBounds` is completely inside `drawableBounds`.

I would replace this with:
```
const inRange = bounds.intersects(drawableBounds);
```

src/RenderWebGL.js
```diff
@@ -1002,10 +1002,19 @@ class RenderWebGL extends EventEmitter {
...
             // default pick list ignores visible and ghosted sprites.
-            if (drawable.getVisible() && drawable.getUniforms().u_ghost !== 0) {
+            if (inRange && drawable.getVisible() && drawable.getUniforms().u_ghost !== 0) {
```
adroitwhiz: You could move the bounds check in here, like so:

```
if (drawable.getVisible() && drawable.getUniforms().u_ghost !== 0) {
    const drawableBounds = drawable.getFastBounds();
    const inRange = bounds.intersects(drawableBounds);
    if (!inRange) return false;
    // then update the matrix and silhouette...
```
This avoids getting the bounds (and hence updating the transform matrix) for drawables which aren't visible anyway.

src/RenderWebGL.js
```diff
@@ -1019,8 +1028,6 @@ class RenderWebGL extends EventEmitter {
         if (candidateIDs.length === 0) {
             return false;
         }
-
-        const bounds = this.clientSpaceToScratchBounds(centerX, centerY, touchWidth, touchHeight);
         if (bounds.left === -Infinity || bounds.bottom === -Infinity) {
```
adroitwhiz: Why wasn't this check moved up to where `bounds` is now defined?

src/RenderWebGL.js
Ryuurock: I used amend push

Ryuurock:
@adroitwhiz

src/RenderWebGL.js
```diff
@@ -1002,10 +1002,18 @@ class RenderWebGL extends EventEmitter {
...
+        if (bounds.left === -Infinity || bounds.bottom === -Infinity) {
+            return false;
+        }
+
         candidateIDs = (candidateIDs || this._drawList).filter(id => {
             const drawable = this._allDrawables[id];
+            const drawableBounds = drawable.getFastBounds();
+            const inRange = bounds.intersects(drawableBounds);
             // default pick list ignores visible and ghosted sprites.
             if (drawable.getVisible() && drawable.getUniforms().u_ghost !== 0) {
```
adroitwhiz, line 1023: Move `drawable.getFastBounds` and `bounds.intersects(drawableBounds)` in here to avoid doing unnecessary work:

```js
if (drawable.getVisible() && drawable.getUniforms().u_ghost !== 0) {
    const drawableBounds = drawable.getFastBounds();
    const inRange = bounds.intersects(drawableBounds);
    if (!inRange) return false;
```

Ryuurock: done.

Issue #427, opened 2019-03-24
Rounding of vector bounding box dimensions causes jittering of sprite


adroitwhiz:
### Expected Behavior

When you modify a vector costume in the Paint Editor, and its bounding box is recalculated, this should not affect the position of the sprite in any way.

### Actual Behavior

Because the paint editor gives SVGs non-integer viewBoxes, but the canvas renderer renders them at integer sizes, and there's tons of code that assumes integer dimensions of textures, there's all kinds of jittering and ugly artifacts.

Fixing this will require proper handling of both non-integral viewBox bounds and rotation centers. A clear separation between a sprite's "rendered" rotation center and bounds and its "logical" rotation center and bounds must be established.

### Steps to Reproduce

[View this project.](https://scratch.mit.edu/projects/277261358/) Note how in the Paint Editor, Scratch Cat doesn't move at all, but on-stage, he jitters, and nearest-neighbor seams are visible:
![Peek 2019-04-24 22-40](https://user-images.githubusercontent.com/25993062/56706189-0ab84000-66e2-11e9-83b1-fc1f8cd7974e.gif)

### Operating System and Browser

Firefox 65.0.1 and Chromium 71.0.3578.98 on Fedora Linux.

adroitwhiz:
@cwillisf GitHub's pattern matching unintentionally made #527 close this. Mind reopening?

adroitwhiz:
Also see https://scratch.mit.edu/discuss/topic/388506/

BryceLTaylor:
I'm copying this over from #547 which I'm going to close as a duplicate:

The second costume will draw in a slightly different place so that switching between the costumes quickly causes it to "jitter."  

![jitter](https://user-images.githubusercontent.com/1088291/72648375-e32c9600-3948-11ea-8193-ad2554d56fd1.gif)

Here is a project that shows this issue that came from the community: https://scratch.mit.edu/projects/359990077/

Note: sometimes the project won't appear jittery.  I looked at it and saw the jitter but @fsih could not see the jitter.  Going into fullscreen mode removed the jitter.  Returning to regular screen mode again did not put the jitter back.  

Tiny screen mode revealed the jitter every time.

This was reported originally in the forums here: https://scratch.mit.edu/discuss/topic/343787/

### Steps to Reproduce

Create a new project
Open the costume editor for the cat
Delete the second costume for the cat
Duplicate the first costume for the cat
Switch to the code tab
Create this stack:
[...]

PR #529, opened 2019-11-28
Expand convex hull bounds by half a pixel per side

### LINKED ISSUE #528: 1x1 sprites will not stamp after a "say" block is executed
### Expected Behavior

The "stamp" block should work consistently

### Actual Behavior

If a block which causes a speech bubble to appear ("say", "think", "ask") is executed before the "stamp" block, sprites whose texture size is 1x1 will not stamp.

Sprites whose texture size is 1 in one dimension but not the other will produce a 1-pixel line.

This probably has something to do with convex hull calculations.

### Steps to Reproduce

https://scratch.mit.edu/projects/348987704/

adroitwhiz:
### Resolves

Resolves #528

### Proposed Changes

This PR expands convex hull bounds by the radius of half a pixel on each side of the bounding box.

### Reason for Changes

Say you have a costume like this:
image

The pixels of said costume can be treated as square color samples taken at the pixels' centers:
image

The convex hull algorithm calculates the hull based on the pixels' centers, not taking area into account:
convhull

As such, bounds calculated from the convex hull need to be enlarged by the "radius" of each pixel.
If pixels were circular, this would be half the "diameter" (aka the drawable's Scratch-space scale), but since they're square, it's actually (√2 / 2) * the diameter.

This will probably change:
- The position of text bubbles (probably not a big deal)
- "If on edge, bounce" behavior (probably a bigger deal)
- Fencing behavior ([shouldn't be using the convex hull anyway](https://github.com/LLK/scratch-render/pull/478))

If the "if on edge, bounce" behavioral change is a compatibility issue, I can redo this once there's architecture in place for [separating a sprite's "logical" bounds and texture bounds](https://github.com/LLK/scratch-render/issues/427), so the old tight bounds can be used for bouncing and the new expanded ones can be used for stamping.

EDIT: The current "if on edge, bounce" behavior is [extremely inconsistent](https://scratch.mit.edu/projects/349149479/) both with itself and 2.0 (you can get this sprite to begin phasing through the walls if you move it into a corner) so this hopefully shouldn't affect compatibility any more than the current behavior already has.

A semi-related issue you may run into when investigating this further is that `getConvexHullPointsForDrawable` does touching tests at the sprite's "logical size", so double-resolution bitmap costumes (in 3.0, that means all of them) are sampled at half their actual resolution.

This means that, incidentally, double-resolution bitmap sprites with odd-numbered resolution dimensions (like the black circle I linked above in 349149479) appear to be properly fenced currently, but will be fenced half a texel too far inwards by this change. This is because their "logical" sizes are their resolutions divided by 2, and we've started rounding that size up. However, bitmap sprites with even-numbered resolution dimensions were fenced too far *outwards*, and this change fixes that.

### DIFF
--- a/src/Rectangle.js
+++ b/src/Rectangle.js
@@ -29,27 +29,35 @@ class Rectangle {
     /**
      * Initialize a Rectangle to the minimum AABB around a set of points.
      * @param {Array<Array<number>>} points Array of [x, y] points.
+     * @param {number} scale The "Scratch-space" area under each point sample.
      */
-    initFromPointsAABB (points) {
+    initFromPointsAABB (points, scale = 0) {
         this.left = Infinity;
         this.right = -Infinity;
         this.top = -Infinity;
         this.bottom = Infinity;
 
+        // Each point is the center of a pixel. However, said pixels each have area.
+        // We can think of the `scale` parameter as the 'diameter' of each area sample centered around each point.
+        // The bounds must be extended by the 'radius' of each sample on every side.
+        // Since nearest-neighbor sampling gives pixels a square area, the radius is sqrt(2) / 2 = 0.707...
+        // at its longest (from the center of the pixel to one of its corners).
+        const halfPixel = 0.7071067811865476 * scale;
+
         for (let i = 0; i < points.length; i++) {
             const x = points[i][0];
             const y = points[i][1];
-            if (x < this.left) {
-                this.left = x;
+            if ((x - halfPixel) < this.left) {
+                this.left = x - halfPixel;
             }
-            if (x > this.right) {
-                this.right = x;
+            if ((x + halfPixel) > this.right) {
+                this.right = x + halfPixel;
             }
-            if (y > this.top) {
-                this.top = y;
+            if ((y + halfPixel) > this.top) {
+                this.top = y + halfPixel;
             }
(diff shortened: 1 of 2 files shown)


src/Rectangle.js
```diff
@@ -29,27 +29,35 @@ class Rectangle {
...
         this.right = -Infinity;
         this.top = -Infinity;
         this.bottom = Infinity;
 
+        // Each point is the center of a pixel. However, said pixels each have area.
+        // We can think of the `scale` parameter as the 'diameter' of each area sample centered around each point.
+        // The bounds must be extended by the 'radius' of each sample on every side.
+        // Since nearest-neighbor sampling gives pixels a square area, the radius is sqrt(2) / 2 = 0.707...
+        // at its longest (from the center of the pixel to one of its corners).
+        const halfPixel = 0.7071067811865476 * scale;
```
cwillisf, line 45:
```suggestion
        const halfPixel = Math.SQRT1_2 * scale;
```

adroitwhiz:
Waiting for all the other convex hull issues (https://github.com/LLK/scratch-render/pull/479 and https://github.com/LLK/scratch-render/issues/592) to be resolved before redoing this

adroitwhiz:
Closing in favor of #724

Issue #543, opened 2020-01-10
The "touching mouse" block doesn't work with vector costumes which have never appeared


adroitwhiz:
### Expected Behavior

The "touching mouse" block should work with all sprites.

### Actual Behavior

The "touching mouse" block only works with vector costumes if they have previously been shown on screen.

### Steps to Reproduce

See [this project](https://scratch.mit.edu/projects/358459259/)

### Operating System and Browser

All

BryceLTaylor:
That project does not indicate a bug.  The second sprite is hidden, so it will not interact with collisions.  Also, the block is "touching mouse-pointer" and not "touching sprite2"

adroitwhiz:
I've deleted the other two sprites to make the problem clearer. When sprite1's costume is costume3, "touching mouse" should return true if the mouse is in the area covered by costume3. This doesn't happen if costume3 has never been rendered (in the demo project, the sprite switches back to costume4 before each frame is rendered).

BryceLTaylor:
So touching sprites works fine, which implies a difference between touching sprite v. touching mouse pointer.  It could be when collisions are calculated, or how the data is stored about what to check.  Or something else.

adroitwhiz:
It's because the "touching" data for a vector costume is never calculated until the costume's texture is fetched, which #544 fixes.

PR #555, opened 2020-02-05
Handle matrix + silhouette updates better

### LINKED ISSUE #545: drawableScale should consider current render target
### Expected Behavior

The scale of a Skin, including MIP choice for SVGSkin, should depend on the current render target. That is, the value returned by `_getDrawableScreenSpaceScale` should be different depending on whether we're drawing / going to draw to the query buffer or the buffer associated with the visible canvas. The method should probably be renamed along with this change, since it'll no longer always be associated with screen space.

### Actual Behavior

Currently the scaling factor is determined as if we're drawing to the buffer associated with the visible canvas, regardless of the actual current render target. This can lead to choosing a larger rasterization of an SVG skin than necessary, which could lead to minor artifacts in render queries.

I'm currently not aware of any project which works incorrectly due to this issue.

### LINKED ISSUE #546: Follow up on the issue described in #398
As a follow-up to #544, I'm considering moving SVG skin silhouette creation into `updateSilhouette`. To ensure this doesn't cause any breakage, we need to ensure the issue in https://github.com/LLK/scratch-render/pull/398 has been fixed.

Here are all the call sites of `Drawable.isTouching`, as far as I can tell:
- `isTouchingColor` and `isTouchingDrawables` call `_candidatesTouching`, which calls `updateSilhouette` on all "candidate" drawables' skins. However, they do not appear to call `updateSilhouette` on the passed drawable's skin.
- `drawableTouching` and `pick` both call `updateSilhouette`.
- `_getConvexHullPointsForDrawable` calls `drawable.skin.isTouchingLinear` directly, and does **not** call `updateSilhouette`.

It may also be fruitful to investigate whether we can throw an error inside `isTouching` if the silhouette is "dirty" without the extra check slowing things down.

adroitwhiz:
### Resolves

Resolves (part of) #545 
Resolves #546
Resolves https://github.com/LLK/scratch-gui/issues/5647

### Proposed Changes

This PR makes the following changes:
- Add `updateCPURenderAttributes` method to `Drawable`
- Move CPU render attribute update from `RenderWebGL._touchingBounds` to `RenderWebGL.isTouchingColor` and `RenderWebGL.isTouchingDrawables`
- Add documentation notes to `Drawable.isTouching`, `Skin.isTouchingNearest`, and `Skin.isTouchingLinear` explaining that those functions' callers are responsible for updating the silhouette + matrix
- Update CPU render attributes in `RenderWebGL._getConvexHullPointsForDrawable`

### Reason for Changes

- Add `updateCPURenderAttributes` method to `Drawable`
  - There were many places in `RenderWebGL` that called both `Drawable.updateMatrix` and `Drawable.skin.updateSilhouette`, and none that called just one or the other. Some checked whether the drawable's skin existed, others did not. I've consolidated calls to `Drawable.updateMatrix` and `Drawable.skin.updateSilhouette` into a single utility method, for conciseness and consistency.
  - This new consolidated method allows for future optimizations like [this one not included in this PR](https://github.com/adroitwhiz/scratch-render/commit/1f6fa1231713006789b02a7e6610961b871eb424), which in my testing improves "touching sprite" performance by 40-70%.
  - `updateCPURenderAttributes` does *not* multiply the drawable's scale by the screen-space scaling ratio, since CPU rendering always occurs at the "native" stage size.

- ~~Move CPU render attribute update from `RenderWebGL._touchingBounds` to `RenderWebGL.isTouchingColor` and `RenderWebGL.isTouchingDrawables`~~
  - ~~`_touchingBounds` doesn't actually require the passed drawable's matrix or silhouette to be up-to-date. However, `isTouchingColor` and `isTouchingDrawables` relied on it updating those. For clarity (and to prevent things from breaking if a future coder decided to "optimize" `_touchingBounds`), these calls have been moved (in the form of calls to the new `updateCPURenderAttributes` function) to the places that *actually* require the drawable's matrix and silhouette to be updated.~~ Moved to another PR

- Add documentation notes to `Drawable.isTouching`, `Skin.isTouchingNearest`, and `Skin.isTouchingLinear` explaining that those functions' callers are responsible for updating the silhouette + matrix
  - This should make it clearer to future coders that they cannot use a drawable's silhouette without updating it.

- ~~Update CPU render attributes in `RenderWebGL._getConvexHullPointsForDrawable`~~
  - ~~This appears to be what was causing the issue in #398. `_getConvexHullPointsForDrawable` calls methods which require the drawable's matrix + silhouette to be up-to-date, but previously did not update them.~~ Done in a previous PR

I have tested this manually and confirmed that it does *not* regress #543.

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -105,7 +105,7 @@ class SVGSkin extends Skin {
         return mip;
     }
 
-    updateSilhouette (scale = 1) {
+    updateSilhouette (scale = [100, 100]) {
         // Ensure a silhouette exists.
         this.getTexture(scale);
     }
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -5,6 +5,7 @@ const RenderConstants = require('./RenderConstants');
 const ShaderManager = require('./ShaderManager');
 const Skin = require('./Skin');
 const EffectTransform = require('./EffectTransform');
+const log = require('./util/log');
 
 /**
  * An internal workspace for calculating texture locations from world vectors
@@ -640,7 +641,11 @@ class Drawable {
     updateCPURenderAttributes () {
         this.updateMatrix();
         // CPU rendering always occurs at the "native" size, so no need to scale up this._scale
-        if (this.skin) this.skin.updateSilhouette(this._scale);
+        if (this.skin) {
+            this.skin.updateSilhouette(this._scale);
+        } else {
+            log.warn(`Could not find skin for drawable with id: ${this._id}`);
+        }
     }
 
     /**
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -452,6 +452,8 @@ class Drawable {
 
     /**
      * Check if the world position touches the skin.
+     * The caller is responsible for ensuring this drawable's inverse matrix & its skin's silhouette are up-to-date.
+     * @see updateCPURenderAttributes
      * @param {twgl.v3} vec World coordinate vector.
      * @return {boolean} True if the world position touches the skin.
      */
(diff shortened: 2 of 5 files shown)


src/SVGSkin.js
```diff
@@ -104,7 +104,7 @@ class SVGSkin extends Skin {
         return mip;
     }
 
-    updateSilhouette (scale = 1) {
+    updateSilhouette (scale = [100, 100]) {
```
adroitwhiz, line 108: :flushed: 

(fortunately, this didn't break anything--all call sites specified a scale explicitly)

PR #556, opened 2020-02-05
Un-premultiply extracted drawables

### LINKED ISSUE #551: Dark outline around dragged sprites
### Expected Behavior

Sprites should have the same appearance when being dragged as when they're not being dragged

### Actual Behavior

If you drag a sprite in the editor, a dark outline appears around it:
image

Note that this was *not* introduced by #515--as far as I can tell, it's always been a problem.

The problem is that `extractDrawable` returns data from `gl.readPixels`, which is premultiplied (and has been even before #515, which only moved the premultiplication out of the shader and into the texture upload). All DOM APIs which consume image data, on the other hand, expect non-premultiplied ("straight") alpha.

My first premultiplication PR (#460) added a new draw mode which "un-premultiplies" the alpha of drawn sprites, which I left out of the second in order to reduce the scope of the changes. I'll probably end up making a new PR with said "un-premultiplying" draw mode.

### Steps to Reproduce

1. Create a lightly-colored sprite
2. Drag it

adroitwhiz:
### Resolves

Resolves #551

### Proposed Changes

This PR adds a new `straightAlpha` draw mode which un-premultiplies rendered sprites by dividing their RGB channels by their alpha channel in the shader, and uses that draw mode in `RenderWebGL.extractDrawable`.

### Reason for Changes

`RenderWebGL.extractDrawable` returns data which is expected to be un-premultiplied, as `scratch-gui` places it into an `ImageData` object and draws it to a canvas. If the data is premultiplied, partially-transparent pixels will appear darker than they should.

### DIFF
--- a/src/ShaderManager.js
+++ b/src/ShaderManager.js
@@ -154,12 +154,12 @@ ShaderManager.EFFECTS = Object.keys(ShaderManager.EFFECT_INFO);
  */
 ShaderManager.DRAW_MODE = {
     /**
-     * Draw normally.
+     * Draw normally. Its output will use premultiplied alpha.
      */
     default: 'default',
 
     /**
-     * Un-premultiply alpha. Useful for reading pixels from GL into an ImageData object.
+     * Draw with non-premultiplied alpha. Useful for reading pixels from GL into an ImageData object.
      */
     straightAlpha: 'straightAlpha',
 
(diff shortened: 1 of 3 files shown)


src/ShaderManager.js
```diff
@@ -158,6 +158,11 @@ ShaderManager.DRAW_MODE = {
      */
     default: 'default',
```
cwillisf: Just to clarify the situation, could you add to the comment on the default draw mode to indicate that its output uses premultiplied alpha?

PR #557, opened 2020-02-20
Clean up PenSkin code

### LINKED ISSUE #450: PenSkin still has a canvas, which it never uses
### Expected Behavior

The `PenSkin` class, which is now implemented in WebGL, should not create an instance of a `<canvas>` element.

### Actual Behavior

It does, and this `<canvas>` element, despite never being drawn to, is somewhat intertwined into the `PenSkin`'s code.

adroitwhiz:
### Resolves

Resolves #450

### Proposed Changes

This PR removes ~270 lines of unnecessary/obsolete code from `PenSkin`, greatly simplifying it and possibly improving performance:
- Remove `_canvas`, `drawStamp`, `_setAttributes`
  - All pen drawing is now done in GL, so there is no need for a 2D canvas context to draw things on, or functions to draw things to it.
  - Without a canvas context to call `createImageData` on (to create silhouette data), the `ImageData` constructor is used instead. Note that MDN lists support for said constructor in Safari as "unknown". I've confirmed that the `ImageData` constructor works in iOS Safari 13, but I haven't tested on Safari 11. [However, this constructor is already used for video in the VM](https://github.com/LLK/scratch-vm/blob/9b7a9770d70e167ffed887568c6a75a39a812e50/src/io/video.js#L150), so this isn't its first use in the codebase.
  - Instead of getting/setting the canvas' size in places that require it, a `_size` property has been added to `PenSkin`.
- Remove `_bounds`
  - This can also be replaced with `_size`.
- Remove `_silhouetteBuffer`
  - It turns out this was never actually used.
- Remove `_exportTexture` and `_drawToBuffer`
  - Instead of going through the trouble of drawing the framebuffer texture to `_exportTexture` when `getTexture` is called, just expose the framebuffer texture as the skin's texture.
  - Note that `updateSilhouette` now directly sets the framebuffer, instead of going through a draw region that unsets it afterwards. This behavior is already present in `clear`, so it shouldn't cause new problems.
- Remove `_stampShader` + `_drawRectangle` and its draw region
  - These were only used by `_enterDrawToBuffer` and `_drawToBuffer`.
- Remove `__modelTranslationMatrix`, `__modelScalingMatrix`, `__modelMatrix`, `__modelTranslationVector`, `__modelScalingVector`
  - These were only used by `_drawRectangle`.
- Remove `__projectionMatrix`
  - Similarly, only the rectangle-drawing code used this. It was also passed as a uniform to the line-drawing code, but was unused.
- Pass `u_stageSize` in `_enterDrawLineOnBuffer` instead of `_drawLineOnBuffer`
  - This avoids passing the same uniform in each time.
- Inline `_createLineGeometry`
  - Since it's only called once (in the `PenSkin` constructor), I believe inlining it makes the code more clear--you don't have to scroll around as much to see what really happens in the constructor.
- Remove `DefaultPenAttributes`
  - This is in a separate commit because it may be a bit riskier, but [the VM should always provide default pen attributes](https://github.com/LLK/scratch-vm/blob/develop/src/extensions/scratch3_pen/index.js#L147), making `DefaultPenAttributes` redundant.

### Reason for Changes

These changes improve ergonomics and make the code easier to improve in the future. They may also increase performance.

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -68,6 +68,12 @@ class PenSkin extends Skin {
             exit: () => this._exitDrawLineOnBuffer()
         };
 
+        /** @type {object} */
+        this._usePenBufferDrawRegionId = {
+            enter: () => this._enterUsePenBuffer(),
+            exit: () => this._exitUsePenBuffer()
+        };
+
         /** @type {twgl.BufferInfo} */
         this._lineBufferInfo = twgl.createBufferInfoFromArrays(this._renderer.gl, {
             a_position: {
@@ -130,10 +136,10 @@ class PenSkin extends Skin {
      * Clear the pen layer.
      */
     clear () {
-        const gl = this._renderer.gl;
-        twgl.bindFramebufferInfo(gl, this._framebuffer);
+        this._renderer.enterDrawRegion(this._usePenBufferDrawRegionId);
 
         /* Reset framebuffer to transparent black */
+        const gl = this._renderer.gl;
         gl.clearColor(0, 0, 0, 0);
         gl.clear(gl.COLOR_BUFFER_BIT);
 
@@ -202,6 +208,20 @@ class PenSkin extends Skin {
         const gl = this._renderer.gl;
 
         twgl.bindFramebufferInfo(gl, null);
+    }
+
+    /**
+     * Prepare to do things with this PenSkin's framebuffer
+     */
+    _enterUsePenBuffer () {
+        twgl.bindFramebufferInfo(this._renderer.gl, this._framebuffer);
+    }
+
+    /**
+     * Return to a base state
+     */
+    _exitUsePenBuffer () {
+        twgl.bindFramebufferInfo(this._renderer.gl, null);
     }
 
     /**
@@ -311,9 +331,9 @@ class PenSkin extends Skin {
      */
     updateSilhouette () {
         if (this._silhouetteDirty) {
+            this._renderer.enterDrawRegion(this._usePenBufferDrawRegionId);
             // Sample the framebuffer's pixels into the silhouette instance
             const gl = this._renderer.gl;
-            twgl.bindFramebufferInfo(gl, this._framebuffer);
             gl.readPixels(
                 0, 0,
                 this._size[0], this._size[1],
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -3,7 +3,6 @@ const twgl = require('twgl.js');
 const RenderConstants = require('./RenderConstants');
 const Skin = require('./Skin');
 
-const Rectangle = require('./Rectangle');
 const ShaderManager = require('./ShaderManager');
 
 /**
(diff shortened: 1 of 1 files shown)


src/PenSkin.js
```diff
@@ -218,7 +149,7 @@ class PenSkin extends Skin {
     drawLine (penAttributes, x0, y0, x1, y1) {
         // For compatibility with Scratch 2.0, offset pen lines of width 1 and 3 so they're pixel-aligned.
         // See https://github.com/LLK/scratch-render/pull/314
-        const diameter = penAttributes.diameter || DefaultPenAttributes.diameter;
```
fsih: It looks like at https://github.com/LLK/scratch-vm/blob/9329f93e74ffbb924b12f318a41db69e1387db74/src/extensions/scratch3_pen/index.js#L188 we expect penState to have a penAttributes attribute, otherwise it will crash, which isn't guaranteed when pen state is set

It also seems like when pen state is set, we don't guarantee that penAttributes properly has the .diameter and .color4f attributes, which I guess is why DefaultPenAttributes existed. Seems like where customState is set, we should be verifying that the format of the object is correct in order to safely remove this safety check (that's all in VM though)

src/PenSkin.js
```diff
@@ -516,66 +275,36 @@ class PenSkin extends Skin {
...
     updateSilhouette () {
         if (this._silhouetteDirty) {
-            if (this._canvasDirty) {
-                this._drawToBuffer();
-            }
-
-            // Render export texture to another framebuffer
-            const gl = this._renderer.gl;
-
-            this._renderer.enterDrawRegion(this._toBufferDrawRegionId);
```
fsih, line 573: We want to make sure we are always using draw regions for consistency/protection, so rather than changing this to be like clear (by moving bindFrameBufferInfo outside of the enterDrawRegion function), we'd rather change clear to call enterDrawRegion.

Can you explain the logic behind removing the `this._drawRectangleRegionEnter(this._stampShader, this._bounds);` step, which used t be called by enterDrawRegion? Are those always guaranteed to be no-ops? What makes sure that state is set?

adroitwhiz: It looks like `_drawRectangleRegionEnter` sets the viewport and shader.

Calling `clear` on the `PenSkin` just calls `gl.clear`, which requires the framebuffer to be set but does *not* depend on the current viewport or shader.

Likewise for `updateSilhouette` which calls `gl.readPixels`, which also only requires the correct framebuffer to be bound.

The code for stamping lives in `RenderWebGL` and never called this code so it shouldn't be affected--it uses the same code path as drawing to the stage and sets the viewport + framebuffer itself.

Finally, the code for pen lines uses an existing draw region which sets the viewport, framebuffer, and shader.

src/PenSkin.js
adroitwhiz: I'll just leave out the default pen attributes changes for now until that can be revisited in the VM

cwillisf (approved):
👍 thanks for the explanations (and your patience...)

fsih (approved):
LG, thanks!

PR #559, opened 2020-02-28
Revert "Draw pen lines via fragment shader"

fsih:
Reverts LLK/scratch-render#438
Bryce found some issues with pen that we're investigating
screen_shot_2020-02-28_at_9 34 29_am

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -44,6 +44,11 @@ const __projectionMatrix = twgl.m4.identity();
  */
 const __modelTranslationMatrix = twgl.m4.identity();
 
+/**
+ * Reused memory location for rotation matrix for building a model matrix.
+ * @type {FloatArray}
+ */
+const __modelRotationMatrix = twgl.m4.identity();
 
 /**
  * Reused memory location for scaling matrix for building a model matrix.
@@ -124,7 +129,7 @@ class PenSkin extends Skin {
         this._stampShader = this._renderer._shaderManager.getShader(ShaderManager.DRAW_MODE.default, NO_EFFECTS);
 
         /** @type {twgl.ProgramInfo} */
-        this._lineShader = this._renderer._shaderManager.getShader(ShaderManager.DRAW_MODE.line, NO_EFFECTS);
+        this._lineShader = this._renderer._shaderManager.getShader(ShaderManager.DRAW_MODE.lineSample, NO_EFFECTS);
 
         this._createLineGeometry();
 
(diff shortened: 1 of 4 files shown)


adroitwhiz:
Keep me updated! What test project is that?

fsih:
I think this is go to random position? I don't think it was doing that when I tested it

adroitwhiz:
This looks more like fencing behavior to me. I'm not sure what problem @BryceLTaylor is referring to in the above screenshot, but I can reproduce something similar-looking on [this project](https://scratch.mit.edu/projects/372002485/):
image

Of course, the sprite's rotation center is waaaay off, so it's to be expected that the top and left sides of the stage aren't getting touched:
image

It doesn't seem like the pen skin is getting incorrectly moved or the lines aren't being drawn in the right place-- all the lines are connected, and they never go off the right or bottom edges of the stage.

The only other issue I see in that screenshot is that some of the pen lines look "dashed":
image

but that's a problem with nearest-neighbor interpolation that existed long before my PR was merged.

fsih:
That makes sense about the fencing behavior.

Do you happen to know what might cause the "problem with nearest-neighbor interpolation"? I agree it seems unlikely it's your PR.

adroitwhiz:
Sorry--by that I meant that the problem *is* nearest-neighbor interpolation. In "smallish stage mode", nearest-neighbor interpolation causes certain rows and columns of pixels to disappear entirely, creating the appearance of dashed lines.

I can think of three ways to fix this:
- The first way is to set `PenSkin.isRaster` to `false`. This will, however, make the pen layer appear blurry at larger sizes (currently it appears pixelated, and this matches 2.0).
- The second way is to modify `useNearest` to use nearest-neighbor interpolation only when scaling *up* raster skins, and to use linear interpolation when scaling them down. I think this actually more closely matches 2.0's behavior, but would change the way bitmap skins are currently displayed.
- The third way would be to add a special case to `useNearest` for when the drawable's skin is a `PenSkin`, or give each `Skin` its own `useNearest` implementation that `Drawable.useNearest` defers to, and implement the "nearest neighbor when scaling up" behavior only for `PenSkin`s.

BryceLTaylor:
This is the project on production that was having the fencing issue: https://scratch.mit.edu/projects/371983393/
I made it on staging first where it exhibited the fencing bug, then I downloaded it from there and uploaded it to production.  The project was not changed between the two versions and it shows no issues on production.

It also demonstrated the skinny lines appearing to be dashed problem, which even if it's caused by existing issues, we should not implement anything that reveals it.

adroitwhiz:
Are you sure the fencing bug is related to my PR? I ran your test project under the "pen shader" branch and found no such issue:
image
(Firefox 73.0.1 on Linux; what browser did you test on?)

Regarding pen lines appearing "dashed", IMO they appear only slightly worse in my PR--here's what they look like currently:
image

I'd prefer to avoid blocking this on a nearest-neighbor fix, as I'm concerned it would take another few months to get that merged in (which is completely understandable given your current workload, but something I'd rather not hold up this PR on). Once a Scratch Team member can provide input on [which of these three options](https://github.com/LLK/scratch-render/pull/559#issuecomment-592760711) should be implemented in order to fix the scaling artifacts, I'd be happy to create another PR.

adroitwhiz:
Alright @BryceLTaylor, I think I *may* have found the source of the fencing issue.

If you draw an off-center costume, the rotation center and bounds are updated as expected and fencing occurs.

If, however, you then *delete the contents of the costume*, fencing incorrectly continues to be applied using the *previous costume contents* until you reload:
Peek 2020-03-04 23-45

PR #561, opened 2020-03-05
Set rotation center to [0, 0] in setEmptyImageData

### LINKED ISSUE #562: Emptied-out costumes retain their previous rotation centers
### Expected Behavior

The rotation center of an empty costume should be (0, 0).

### Actual Behavior

When the contents of a costume are deleted in the paint editor, the now-empty costume's rotation center will still be set to whatever it was previously, only being set to (0, 0) upon a reload. This can cause fencing issues.

See [here](https://github.com/LLK/scratch-render/pull/559#issuecomment-595026465) for an instance of this.

### Steps to Reproduce

1. Open the paint editor
2. Delete the contents of the current costume
3. Observe how far the sprite is now fenced from the edges

N/A

adroitwhiz:
### Resolves

Resolves #562 

### Proposed Changes

This PR adds logic in `Skin.setEmptyImageData` to set the `Skin`'s rotation center to `[0, 0]`.

### Reason for Changes

Both [`BitmapSkin`](https://github.com/LLK/scratch-render/blob/a28753fb5a901073f7b61c2ada7505d183cb2b83/src/BitmapSkin.js#L79) and [`SVGSkin`](https://github.com/LLK/scratch-render/blob/develop/src/SVGSkin.js#L164) have logic to call `setEmptyImageData` and return early if they are passed an image whose width or height are 0.

However, because of this early return, their rotation centers are never properly set, and so continue to use their previously set values. The first time you load a project, these will be initialized to `[0, 0]`, but when you, for instance, delete the contents of a previously non-empty costume, they will be set to the previous costume's rotation center. This causes the costume to be fenced in too far.

I'm directly setting the skin's `_rotationCenter` instead of calling `setRotationCenter` both to avoid emitting `Skin.Events.WasAltered` twice, and because [I plan to remove `setRotationCenter` eventually](https://github.com/LLK/scratch-vm/pull/2314).

### DIFF
--- a/src/Skin.js
+++ b/src/Skin.js
@@ -212,6 +212,9 @@ class Skin extends EventEmitter {
             this._emptyImageTexture = twgl.createTexture(gl, textureOptions);
         }
 
+        this._rotationCenter[0] = 0;
+        this._rotationCenter[1] = 0;
+
         this._silhouette.update(this._emptyImageData);
         this.emit(Skin.Events.WasAltered);
     }


fsih:
We are in the process of updating our contributor guidelines to be consistent across repos and more useful. One of the rules we're planning to be stricter about again is that we want all PRs to have issues associated with them. This way, if it's an issue we aren't planning to fix, or one that requires discussion or design from the scratch team, we can close it or add clarification to it before a contributor puts a lot of work into a PR that might end up being turned away. I know the problem in the past has been that people have filed issues and they've been ignored 😓and we're working on having a process where we queue them all up to be reviewed.

With that in mind, could you file an issue for this PR, and I'll assign it back to you?

adroitwhiz:
An issue has been created: #562 

Thanks for taking the time to review these PRs!

It may be beneficial to clarify the "associated issue" thing in the pull request template, maybe something like:
```md
### Resolves

_Include a link to the GitHub issue that this PR resolves. If an issue does not exist, please create one._
```

adroitwhiz:
Also, where would performance improvement PRs fit into this? Oftentimes, the "issue" for those isn't "this operation is particularly slow", just "this operation could be faster", and explaining *why* it could be faster essentially requires you to do the optimization in the first place.

fsih:
@adroitwhiz that's a great question! Performance is usually quite difficult to contribute to because it usually requires a deep understanding the whole system, so we weren't really thinking about as we were rewriting our contributor guidelines, which are mostly aimed at newbies. You're right that it would be pretty difficult to predict how much success you would have before you started trying stuff.

We have a performance label in github, so you can see examples of issues we've written for ourselves for performance investigation: https://github.com/LLK/scratch-render/issues?q=is%3Aissue+label%3Aperformance They're usually along the lines of, this looks like it could potentially benefit from investigation, should we do it? We would probably have to go with our and your best guess on which of those kinds of issues would be most worth pursuing.

PR #566, opened 2020-03-12
Move useNearest from Drawable to Skin

### LINKED ISSUE #565: Pen lines appear dashed in small stage mode
### Expected Behavior

Pen lines should never appear "dashed" or broken-up.

### Actual Behavior

In small-stage mode and "smallish-stage mode" (which occurs when the browser window is too small to fit a normal-sized stage), pen lines appear dashed and especially jagged, and some segments go missing:
image
image

This is a consequence of the nearest-neighbor texture filtering applied to the `PenSkin`, which downscales textures by simply removing some rows and columns of pixels, instead of blending them with neighboring pixels as linear texture filtering does.

### Steps to Reproduce

1. Enter small-stage mode (or enter "smallish-stage mode" by resizing your browser window until the stage shrinks slightly)
2. Draw some pen lines at pen size 1
3. Observe the lines' appearance

All

adroitwhiz:
### Resolves

Resolves #565 

### Proposed Changes

This PR moves the `useNearest` method from the `Drawable` class to the `Skin` class. The default behavior is to always use nearest-neighbor interpolation, but `Skin` subclasses can override the method to provide their own behavior, and two currently do:
- `SVGSkin.useNearest` has absorbed the complex logic previously present in `Drawable.useNearest` (checking if certain effects are set, the sprite is rotated, and the scale is close to 100%).
- `PenSkin.useNearest` will use nearest-neighbor interpolation when being scaled up, and linear interpolation when being scaled down. This prevents pen lines from appearing "dashed" when a `PenSkin` is displayed at <100% scale.

### Reason for Changes

I believe this is the clearest way to implement the desired interpolation behavior for `PenSkin`s.

Since textures belong to a `Skin`, it makes sense that deciding the proper way to interpolate a texture should be the responsibility of that `Skin`.

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -59,8 +59,6 @@ class SVGSkin extends Skin {
         return this._svgRenderer.size;
     }
 
-    // Because SVG skins' bounding boxes are currently not pixel-aligned, the idea here is to hide blurriness
-    // by using nearest-neighbor scaling if one screen-space pixel is "close enough" to one texture pixel.
     useNearest (scale) {
         // If the effect bits for mosaic, pixelate, whirl, or fisheye are set, use linear
         if ((this.enabledEffects & (
@@ -77,8 +75,12 @@ class SVGSkin extends Skin {
             return false;
         }
 
+        // Because SVG skins' bounding boxes are currently not pixel-aligned, the idea here is to hide blurriness
+        // by using nearest-neighbor scaling if one screen-space pixel is "close enough" to one texture pixel.
         // If the scale of the skin is very close to 100 (0.99999 variance is okay I guess)
-        // TODO: Make this check more precise. Mipmaps complicate this somewhat.
+        // TODO: Make this check more precise. We should use nearest if there's less than one pixel's difference
+        // between the screen-space and texture-space sizes of the skin. Mipmaps make this harder because there are
+        // multiple textures (and hence multiple texture spaces) and we need to know which one to choose.
         if (Math.abs(scale[0]) > 99 && Math.abs(scale[0]) < 101 &&
             Math.abs(scale[1]) > 99 && Math.abs(scale[1]) < 101) {
             return true;
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -2,6 +2,7 @@ const twgl = require('twgl.js');
 
 const Skin = require('./Skin');
 const SvgRenderer = require('scratch-svg-renderer').SVGRenderer;
+const ShaderManager = require('./ShaderManager');
 
 const MAX_TEXTURE_DIMENSION = 2048;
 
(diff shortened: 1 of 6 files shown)


fsih:
Could you resolve conflicts?

adroitwhiz:
Rebased to latest develop

src/SVGSkin.js
```diff
@@ -58,6 +59,33 @@ class SVGSkin extends Skin {
         return this._svgRenderer.size;
     }
 
+    // Because SVG skins' bounding boxes are currently not pixel-aligned, the idea here is to hide blurriness
```
fsih: Move this comment down t line 80?

src/SVGSkin.js
```diff
@@ -58,6 +59,33 @@ class SVGSkin extends Skin {
...
+            return false;
+        }
+
+        // We can't use nearest neighbor unless we are a multiple of 90 rotation
+        if (this._direction % 90 !== 0) {
+            return false;
+        }
+
+        // If the scale of the skin is very close to 100 (0.99999 variance is okay I guess)
+        // TODO: Make this check more precise. Mipmaps complicate this somewhat.
```
fsih: Clarify this TODO in case someone else has to implement it

fsih (dismissed):
LGTM!

adroitwhiz:
For some reason this is making vector sprites look blurry

fsih:
Sorry this fell off our radar :/
It looks like there's a somewhat tricky conflict here, do you mind resolving it?

src/SVGSkin.js
```diff
@@ -58,6 +59,35 @@ class SVGSkin extends Skin {
...
+            ShaderManager.EFFECT_INFO.pixelate.mask |
+            ShaderManager.EFFECT_INFO.mosaic.mask
+        )) !== 0) {
+            return false;
+        }
+
+        // We can't use nearest neighbor unless we are a multiple of 90 rotation
+        if (drawable._direction % 90 !== 0) {
+            return false;
+        }
```
fsih, line 76: What do you think of keeping the drawable.useNearest function, having these lines stay there, and then having it return skin.useNearest? Then both drawable and skin get to have a say in whether we useNearest

adroitwhiz: These lines are very `SVGSkin`-specific--it kinda sucks that `SVGSkin` depends on its parent drawable to know whether it should use nearest-neighbor interpolation but I'm not sure what else can be done.

cwillisf (approved):
👍 thanks!

There are some weird things going on here -- such as https://github.com/LLK/scratch-render/pull/566/files#r510419973 -- but after some thinking and discussing I think I agree that the weirdness is sort of inherent to the way it's all set up and isn't a problem introduced by this change. Maybe we can find a way to clean it up later, but I'm happy to move forward.

Issue #294, opened 2018-05-25
Blurry when dragging - Extract drawable doesn't respect pixel ratio


paulkaplan:
### Expected Behavior

_Please describe what should happen_

`extractDrawable` should return image data that is appropriately scaled for the current devicePixelRatio. 

### Actual Behavior

_Describe what actually happens_

Currently it always returns imagedata rendered at 1x, causing things that use `extractDrawable` to be blurry on hi dpi screens (e.g. editor-style dragging in scratch-gui)

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

Using a hi dpi screen, see how extractDrawable looks, either in production (by dragging a sprite on https://llk.github.io/scratch-gui/develop) or the playground.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

Hi DPI screen, tested on modern retina macbook.

---

/cc @cwillisf. I'm not sure what you think the best way to deal with this is. I was able to fix it pretty simply by scaling the drawable bounds width/height in the calls to `gl.viewport`, the creation of the imageData, the `gl.readPixels` call and the width/height properties of the returned object. The "scale" of the returned imageData should probably be included with the `DrawableExtraction` object. Does that sound reasonable?

evhan55:
Can confirm I see this too on MacBook Pro (Retina, 15-inch, Mid 2015).

Whenever https://github.com/LLK/scratch-render/pull/376 I can try to look at it since I was just working in that area.

PR #568, opened 2020-03-12
Add extractDrawableScreenSpace method

adroitwhiz:
~~This PR depends on #556~~ (merged now)

### Resolves

A step towards resolving https://github.com/LLK/scratch-render/issues/294, https://github.com/LLK/scratch-gui/issues/2388, https://github.com/LLK/scratch-gui/issues/1782, https://github.com/LLK/scratch-gui/issues/1894

### Proposed Changes

This PR deprecates `RenderWebGL.extractDrawable` in favor of a new method, `RenderWebGL.extractDrawableScreenSpace`. This new method differs in several ways:
- Most obviously, it functions in screen-space: it draws the drawable pixels at the same resolution as the regular canvas.
- It returns `ImageData` instead of a `Uint8Array`-- since this method is only used in the GUI once, that call site creates an `ImageData` from the result anyway, and we need to return the data's width and height somehow, it makes sense to just return an `ImageData` which has its width and height built in.
- It does not take `x` and `y` parameters--the math it did with those can be moved entirely to the GUI.
- It explicitly throws an error when the corresponding drawable for the passed `drawableID` does not exist. Previously it returned `null` and this was not handled by the GUI. An explicit error should result in a much more friendly and searchable error message if this function ever fails in the future.
- It does not add depth or stencil attachments to the framebuffer it creates.
- It deletes the framebuffer it creates, preventing memory leaks.
- ~~It applies the ghost effect to the extracted drawable, like 2.0 did when dragging sprites (this is why this PR depends on #556 -- ghosted sprites will appear too dark if the alpha is not properly un-premultiplied).~~ saving this for another day
- The one I'm the least sure about: ~~it doesn't attempt to handle WebGL operations failing/erroring out in various ways that may or may not be handled throughout the rest of the codebase. I'd be hesitant to attempt to add error-handling code for errors I cannot reproduce or trigger in some ways, but I'd like your opinions and reasoning behind why the previous `extractDrawable` handled errors the way it did.~~ I've added a `finally` clause that ensures that the created framebuffer is always deleted. I haven't added a fallback to `queryBufferInfo` just in case the created framebuffer is incomplete-- under what circumstances does that occur?

### Reason for Changes

These changes are a step towards resolving several sprite-dragging bugs.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1242,7 +1242,8 @@ class RenderWebGL extends EventEmitter {
 
             gl.clearColor(0, 0, 0, 0);
             gl.clear(gl.COLOR_BUFFER_BIT);
-            this._drawThese([drawableID], ShaderManager.DRAW_MODE.straightAlpha, projection);
+            this._drawThese([drawableID], ShaderManager.DRAW_MODE.straightAlpha, projection,
+                {effectMask: ~ShaderManager.EFFECT_INFO.ghost.mask});
 
             const data = new Uint8Array(Math.floor(clampedWidth * clampedHeight * 4));
             gl.readPixels(0, 0, clampedWidth, clampedHeight, gl.RGBA, gl.UNSIGNED_BYTE, data);
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1196,8 +1196,11 @@ class RenderWebGL extends EventEmitter {
         const scratchBounds = drawable.getFastBounds();
 
         const canvas = this.canvas;
+        // Ratio of the screen-space scale of the stage's canvas to the "native size" of the stage
         const scaleFactor = canvas.width / this._nativeSize[0];
 
+        // Bounds of the extracted drawable, in "canvas pixel space"
+        // (origin is 0, 0, destination is the canvas width, height).
         const canvasSpaceBounds = new Rectangle();
         canvasSpaceBounds.initFromBounds(
             (scratchBounds.left + nativeCenterX) * scaleFactor,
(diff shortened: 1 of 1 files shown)


fsih:
"It applies the ghost effect to the extracted drawable, like 2.0 did when dragging sprites (this is why this PR depends on #556 -- ghosted sprites will appear too dark if the alpha is not properly un-premultiplied)." - we may want a design pass on this, since we unghosted sprites when dragging in 2.0

src/RenderWebGL.js
```diff
@@ -1163,6 +1170,104 @@ class RenderWebGL extends EventEmitter {
...
+
+            const data = new Uint8Array(Math.floor(clampedWidth * clampedHeight * 4));
+            gl.readPixels(0, 0, clampedWidth, clampedHeight, gl.RGBA, gl.UNSIGNED_BYTE, data);
+            // readPixels can only read into a Uint8Array, but ImageData has to take a Uint8ClampedArray.
+            // We can share the same underlying buffer between them to avoid having to copy any data.
+            const imageData = new ImageData(new Uint8ClampedArray(data.buffer), clampedWidth, clampedHeight);
+
+            // On high-DPI devices, the canvas' width (in canvas pixels) will be larger than its width in CSS pixels.
+            // We want to return the CSS-space bounds,
+            // so take into account the ratio between the canvas' pixel dimensions and its layout dimensions.
```
cwillisf, line 1262: Could you explain in this comment block the relationship between `ratio` here and `window.devicePixelRatio`, and why we need to calculate this explicitly?

src/RenderWebGL.js
```diff
@@ -1163,6 +1170,104 @@ class RenderWebGL extends EventEmitter {
...
+                scratchBounds.right,
+                scratchBounds.top,
+                scratchBounds.bottom,
+                -1, 1
+            );
+
+            gl.clearColor(0, 0, 0, 0);
+            gl.clear(gl.COLOR_BUFFER_BIT);
+            this._drawThese([drawableID], ShaderManager.DRAW_MODE.straightAlpha, projection,
+                {effectMask: ~ShaderManager.EFFECT_INFO.ghost.mask});
```
cwillisf, line 1252: might be worth explicitly commenting about disallowing the ghost effect so the user can see what they're dragging

PR #572, opened 2020-03-19
Fix setRotationCenter empty skin detection

### LINKED ISSUE #571: Changing costume to one with zero width or height doesn't update correctly
### Expected Behavior

When the user clicks a costume in the paint editor, the sprite's bounds should always be updated to match the costume.

### Actual Behavior

When the user clicks a costume in the paint editor, the sprite's bounds are updated to match the costume ONLY if the new costume's width and height are both non-zero.

### Steps to Reproduce

1. Create or edit a project which uses the pen extension
2. Make a sprite with at least two costumes:
   - One with a perfectly vertical or horizontal line, which due to #434 will get a 1x1 ImageData object internally.
   - One "normal" costume where width and height are both non-zero
3. Move the rotation center outside the costume for the "normal" costume
4. Activate the "normal" costume and use `pen down` and `go to random position` to see fencing behavior
5. While the project is running, switch to the other costume.

Expected: the sprite (Drawable) is fenced according to the new 1x1 ImageData
Actual: the sprite (Drawable) is fenced according to the previous "good" costume

The problem goes away after using the paint editor to adjust the rotation center of the zero-width / zero-height costume.

Example (thanks @fsih!)

adroitwhiz:
### Resolves

Resolves #571

### Proposed Changes

This PR changes `setRotationCenter`'s empty skin check to return true if either of the skin's dimensions are 0, whereas previously it only returned true if both were.

### Reason for Changes

`SVGSkin.setSVG` and `BitmapSkin.setBitmap` treat a skin as "empty" if *either* of its dimensions are 0. `setRotationCenter` did not, and because the VM calls `setRotationCenter` upon every costume switch, a costume with zero width *or* zero height (but not both) would have its rotation center improperly set.

### DIFF
--- a/src/Skin.js
+++ b/src/Skin.js
@@ -108,7 +108,7 @@ class Skin extends EventEmitter {
      * @fires Skin.event:WasAltered
      */
     setRotationCenter (x, y) {
-        const emptySkin = this.size[0] === 0 && this.size[1] === 0;
+        const emptySkin = this.size[0] === 0 || this.size[1] === 0;
         // Compare a 32 bit x and y value against the stored 32 bit center
         // values.
         const changed = (


adroitwhiz:
Closing because with https://github.com/LLK/scratch-vm/pull/2314 merged, `setRotationCenter` should never be called anymore.

PR #576, opened 2020-03-22
Don't apply ghost effect to "color is touching color" drawable

### LINKED ISSUE #575: "color () is touching ()" block does not work properly on ghosted sprites
### Expected Behavior

When the "color () is touching ()" block is used, the ghost effect of the sprite calling it should not affect the result.

### Actual Behavior

If the sprite calling "color () is touching ()" has a ghost effect applied, the first argument to "color () is touching()" (the color mask to be applied to the sprite) will be incorrectly darkened, meaning that it will not match properly.

This is a regression introduced by #515.

This has the potential to break [the commonly-used "wall sensor" pattern](https://en.scratch-wiki.info/wiki/Wall_Sensors)-- I can't find any links right now, but I often come across projects that apply the ghost effect to a "wall sensor".

### Steps to Reproduce

1. Add a new sprite
2. Drag the new sprite over an orange region of the default cat
3. Select the default cat
4. Drag a "color () is touching ()" block onto the default cat's block workspace
5. Set the first argument to the default cat's orange color via the eyedropper
6. Set the second argument to one of the overlapping sprite's colors
7. Observe that if you click the block, it returns true
8. Set the default cat's ghost effect to a nonzero value
9. Observe that the block no longer returns true when executed

Or see [this repro project](https://scratch.mit.edu/projects/378517279/editor/).

All

adroitwhiz:
### Resolves

Resolves #575

### Proposed Changes

This PR re-adds the `effectMask` parameter to `sampleColor4b` and `transformColor`, and uses it to mask out the ghost effect for the target drawable of `isTouchingColor`.

### Reason for Changes

#515 made it so that the ghost effect multiplies all color channels by the ghost effect, not just the alpha value. Because of this, if the ghost effect is applied to a drawable, its colors will change, making the color mask invalid.

This was already fixed on the GPU path (which is why the tests pass), but I overlooked it for the CPU path.

### Test Coverage

Waiting on #537 for CPU/GPU parity tests

### DIFF
--- a/src/Drawable.js
+++ b/src/Drawable.js
@@ -679,9 +679,10 @@ class Drawable {
      * @param {twgl.v3} vec The scratch space [x,y] vector
      * @param {Drawable} drawable The drawable to sample the texture from
      * @param {Uint8ClampedArray} dst The "color4b" representation of the texture at point.
+     * @param {number} [effectMask] A bitmask for which effects to use. Optional.
      * @returns {Uint8ClampedArray} The dst object filled with the color4b
      */
-    static sampleColor4b (vec, drawable, dst) {
+    static sampleColor4b (vec, drawable, dst, effectMask) {
         const localPosition = getLocalPosition(drawable, vec);
         if (localPosition[0] < 0 || localPosition[1] < 0 ||
             localPosition[0] > 1 || localPosition[1] > 1) {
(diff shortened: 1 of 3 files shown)


fsih:
oops beaten to it

cwillisf:
For future reference: we ended up replacing #537 with #604 but CPU/GPU testing is no longer blocked

Issue #570, opened 2020-03-19
Effective bounds of zero-width or zero-height skin/costume are incorrect


cwillisf:
### Expected Behavior

A Skin created for a costume should have the same bounds as that costume.

### Actual Behavior

A Skin created for a Wx0 or 0xH (for any value of W or H) goes through the `setEmptyImageData` path which always leads to a 1x1 image (which is fine) and matching bounds (arguably not fine). The difference can be detected through fencing behavior and probably a few other ways.

### Steps to Reproduce

These steps will work until we fix #434 

1. Create or edit a project which uses the pen extension
2. Create or edit a costume with non-zero width and height. The default cat costume is fine.
3. Add a perfectly horizontal or vertical line
4. Delete everything from the costume EXCEPT for the perfectly horizontal or vertical line.
   - Due to #434, a perfectly horizontal line will be treated as having a height of 0, and a perfectly vertical line will be treated as having a width of 0.
5. Adjust the rotation center so that it is outside the model's bounds.
6. Use `pen down` and `forever { go to random position }` to see the fencing behavior. 

Example (thanks @fsih!):
![](https://user-images.githubusercontent.com/2855464/77103769-56a65e80-69f1-11ea-84f5-65a01f45fd9d.gif)

fsih:
Perhaps costumes should never have 0 width or height

adroitwhiz:
This led me down a bit of an investigation, but the TL;DR is that I have a patch that won't work until https://github.com/LLK/scratch-vm/pull/2314 is merged (and preferably, `setRotationCenter` is removed).

- Empty vector costumes are given a rotation center of [240, 180] (both in 2.0 and in 3.0 due to https://github.com/LLK/scratch-paint/issues/983), so always setting skins' rotation centers to the explicitly-passed ones won't work for 0x0 costumes.
- This is handled [at fence-time in 2.0](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/scratch/ScratchSprite.as#L193), but doing something similar here would be kind of janky in my opinion (probably requiring some rotation-center/rectangle coordinate conversion, and possibly creating GC pressure by creating a new `Rectangle` every time a zero-size sprite is moved).
[...]

PR #581, opened 2020-04-02
Remove setRotationCenter API

### LINKED ISSUE #550: Subpixel SVG costume handling: a writeup
## The Problem

When editing a vector sprite, weird errors appear. Sometimes, columns/rows appear doubled up, and the very edges of a sprite may get cut off.

jaggies

This becomes particularly noticeable if you're using vector sprites in an animation, and you draw several different frames:

jitter_old

Notice how unstable this appears-- the cat is jittering around, and sometimes the very bottom of its feet get cut off.

## The Root Cause

This is due to a fundamental difference between vector and raster graphics: subpixels.

An SVG costume's `viewBox`, which describes its bounds, can have non-integer dimensions. You can, for instance, have a sprite that's 2.5 pixels wide. However, eventually the sprite will need to be rendered to a texture, and textures necessarily have integer dimensions.

That's not the only problem, however. To further complicate matters, costumes can have non-integer rotation centers. These rotation centers shift a sprite over by a sub-pixel amount. Consider this SVG of a 3x3 circle (viewbox in red):

circle-1

Let's say we rasterize this circle as-is:

circle-1-rasterized

This appears fine. However, the center of this circle is at (1.5, 1.5). This will cause problems if we want to render the circle in the center of a, say, 6x6 stage:

circle-1-rasterized-centered

Now, we need to do some resampling of the already-rasterized SVG! If we do nearest-neighbor resampling, we get the "jaggies" shown earlier. If we do linear resampling, the SVG appears quite blurry. Clearly, we need to take care of this *before* rasterization.

## The Theory

What we need to do is:
1. Shift the SVG so that its rotation center lies on integer coordinates:

circle-1-shift-1 circle-1-shift-2

2. Expand the bounding out to integer coordinates. Note that this step also solves the problem of non-integer `viewBox`es, although it's not shown here:

circle-1-shift-3

3. Adjust the rotation center of the sprite passed to `scratch-render` so that it is relative to the new bounding box. For instance, the above circle's rotation center is now (2, 2).

## The Practice (Attempt 1)

Now comes the implementation. We modify [`SVGRenderer.loadSVG`](https://github.com/LLK/scratch-svg-renderer/blob/ec35ed3ee88869b48a0316263bd083e1364609a9/src/svg-renderer.js#L126) to take the costume's intended rotation center as an argument, and before creating the SVG image, apply the necessary transformations to its `viewBox`. Then, we offset the rotation center used by `scratch-render` by adding it to `SVGRenderer.viewOffset`.

Unfortunately, this doesn't work with mipmaps.

If you render the costume at above 100% size, it works fine, but if you render a mipmap *under* the costume's "native" size, then the rasterized costume may once again have a non-integer size or rotation center--for instance, if you rasterize at half the costume's normal size, you're dividing the rotation center and dimensions by 2.

## The Practice (Attempt 2)

What if we instead applied this bounding-box adjustment each time the costume was rasterized? `SVGSkin` would pass the costume's rotation center to the [`draw` method](https://github.com/LLK/scratch-svg-renderer/blob/ec35ed3ee88869b48a0316263bd083e1364609a9/src/svg-renderer.js#L404) rather than the `loadSVG` method, and in [`_drawFromImage`](https://github.com/LLK/scratch-svg-renderer/blob/ec35ed3ee88869b48a0316263bd083e1364609a9/src/svg-renderer.js#L432), the offsetting and resizing would be applied to the *canvas* that the SVG is being rendered to.

This works fine... in Chrome. But to implement this, you need to call `ctx.drawImage` at non-integer coordinates. And in Firefox, the `drawImage` method [won't render SVGs with the proper subpixel offset, instead resampling them and making them blurry](https://bugzilla.mozilla.org/show_bug.cgi?id=1517901).

So the only cross-browser way to offset an SVG by a subpixel amount is to set its `viewBox`, like we did in our first attempt. But how do we do this without breaking mipmaps?

## The Practice (Attempt 3)

This is mostly the same as attempt 1, but with one key difference: instead of changing the `viewBox` and rotation center to be round numbers, we ensure they are multiples of the smallest mipmap's size.

For instance, if the smallest mipmap is 1/8th the SVG's native scale, we ensure that the `viewBox` and rotation center are multiples of 8. This ensures that, even at the smallest mipmap's size, the SVG is rendered at the proper subpixel position. This is the best cross-browser solution I've found.

## The Demo

I have a rough version of this working in my `subpixel-v2` branches of [`scratch-render`](https://github.com/adroitwhiz/scratch-render/commit/5285b1266bfab60edfe5a29fb931cea55c095d5a) and [`scratch-svg-renderer`](https://github.com/adroitwhiz/scratch-svg-renderer/commit/6f36758a36493480a86af2db6478c3079e4c505f). The code is somewhat old (preceding SVG mipmaps, the removal of `getDrawRatio`, and the synchronous `scratch-svg-renderer` API), but it works well enough to eliminate the jitter and jaggies entirely:

jitter_new

## The Remaining Problem

There's a kink that still needs to be worked out, however. Expanding the viewbox out to multiples of *n* pixels will increase the size of costumes' bounds. This isn't a problem for `getFastBounds` (which doesn't return consistent results and doesn't need to) and `getConvexHullPointsForDrawable` (the size of the convex hull won't be affected at all), but it presents a problem for `getAABB`, which is used (or rather, [should be used](https://github.com/LLK/scratch-render/pull/478)) for fencing sprites.

With the viewbox-expanding behavior in place, there will be a split between a costume's "logical bounds" (aka the `viewBox`), and its "texture bounds" (the dimensions of the quadrilateral rendered by WebGL).

In order to calculate both the "logical bounds" (for fencing behavior) and "texture bounds" (for rendering), it looks to me like `Drawable._calculateTransform` will have to be refactored in some way-- by changing it to calculate both the "logical" and "texture" bounds, by splitting it into two different methods which calculate the two different bounds, or in some other (hopefully less complex) way.

EDIT: Distortion effects will also have to change what they consider the "center" of the sprite.

If you have any ideas for avoiding complexity here, I'd greatly appreciate them.

adroitwhiz:
### Resolves

A step on the long-ish path towards fixing #550
Also a step towards fixing #570

### Proposed Changes

This PR removes the API for setting a drawable's rotation center and the `Skin.setRotationCenter` function.

It does *not* remove the API for setting a skin's rotation center when setting its contents.

To fix unit tests which previously called `Skin.setRotationCenter`, it also adds a `rotationCenter` setter (plus corresponding getter) to `MockSkin`, which the unit tests now use.

I'm not expecting this to be merged right away--in case https://github.com/LLK/scratch-vm/pull/2314 causes any regressions in the near future, it'll probably be beneficial to be able to revert that without having to re-add the rotation center API here as well.

### Reason for Changes

In order to fix (no, GitHub, it doesn't close) #550, vector costumes' SVG viewboxes must depend on their rotation centers. Every time the viewbox of an SVG is set, its `` must be updated as well, which is a fundamentally asynchronous operation.

As such, a synchronous API for setting a skin's rotation center after the skin is loaded is incompatible with the changes necessary for #550.

With https://github.com/LLK/scratch-vm/pull/2314 merged, that synchronous API is no longer used and can be removed.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -1338,32 +1338,6 @@ class RenderWebGL extends EventEmitter {
     }
 
     /**
-     * Update a drawable's skin rotation center.
-     * @param {number} drawableID The drawable's id.
-     * @param {Array.<number>} rotationCenter The rotation center for the skin.
-     */
-    updateDrawableRotationCenter (drawableID, rotationCenter) {
-        const drawable = this._allDrawables[drawableID];
-        // TODO: https://github.com/LLK/scratch-vm/issues/2288
-        if (!drawable) return;
-        drawable.skin.setRotationCenter(rotationCenter[0], rotationCenter[1]);
-    }
-
-    /**
-     * Update a drawable's skin and rotation center together.
-     * @param {number} drawableID The drawable's id.
-     * @param {number} skinId The skin to update to.
-     * @param {Array.<number>} rotationCenter The rotation center for the skin.
-     */
-    updateDrawableSkinIdRotationCenter (drawableID, skinId, rotationCenter) {
-        const drawable = this._allDrawables[drawableID];
-        // TODO: https://github.com/LLK/scratch-vm/issues/2288
-        if (!drawable) return;
-        drawable.skin = this._allSkins[skinId];
-        drawable.skin.setRotationCenter(rotationCenter[0], rotationCenter[1]);
-    }
-
-    /**
      * Update a drawable's position.
      * @param {number} drawableID The drawable's id.
      * @param {Array.<number>} position The new position.
@@ -1455,9 +1429,6 @@ class RenderWebGL extends EventEmitter {
         }
         if ('skinId' in properties) {
             this.updateDrawableSkinId(drawableID, properties.skinId);
-        }
-        if ('rotationCenter' in properties) {
-            this.updateDrawableRotationCenter(drawableID, properties.rotationCenter);
         }
         drawable.updateProperties(properties);
     }
--- a/src/Skin.js
+++ b/src/Skin.js
@@ -5,19 +5,6 @@ const twgl = require('twgl.js');
 const RenderConstants = require('./RenderConstants');
 const Silhouette = require('./Silhouette');
 
-/**
- * Truncate a number into what could be stored in a 32 bit floating point value.
- * @param {number} num Number to truncate.
- * @return {number} Truncated value.
- */
-const toFloat32 = (function () {
-    const memory = new Float32Array(1);
-    return function (num) {
-        memory[0] = num;
-        return memory[0];
-    };
-}());
-
 class Skin extends EventEmitter {
     /**
      * Create a Skin, which stores and/or generates textures for use in rendering.
@@ -99,26 +86,6 @@ class Skin extends EventEmitter {
      */
     get size () {
         return [0, 0];
-    }
-
-    /**
-     * Set the origin, in object space, about which this Skin should rotate.
-     * @param {number} x - The x coordinate of the new rotation center.
-     * @param {number} y - The y coordinate of the new rotation center.
-     * @fires Skin.event:WasAltered
-     */
-    setRotationCenter (x, y) {
-        const emptySkin = this.size[0] === 0 && this.size[1] === 0;
-        // Compare a 32 bit x and y value against the stored 32 bit center
-        // values.
-        const changed = (
-            toFloat32(x) !== this._rotationCenter[0] ||
-            toFloat32(y) !== this._rotationCenter[1]);
-        if (!emptySkin && changed) {
-            this._rotationCenter[0] = x;
-            this._rotationCenter[1] = y;
-            this.emit(Skin.Events.WasAltered);
-        }
     }
 
     /**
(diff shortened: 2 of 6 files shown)


src/Skin.js
```diff
@@ -101,26 +88,6 @@ class Skin extends EventEmitter {
...
-     * @fires Skin.event:WasAltered
-     */
-    setRotationCenter (x, y) {
-        const emptySkin = this.size[0] === 0 && this.size[1] === 0;
-        // Compare a 32 bit x and y value against the stored 32 bit center
-        // values.
-        const changed = (
-            toFloat32(x) !== this._rotationCenter[0] ||
-            toFloat32(y) !== this._rotationCenter[1]);
-        if (!emptySkin && changed) {
```
fsih, line 117: Note: this is replaced by https://github.com/LLK/scratch-render/pull/561/files

src/RenderWebGL.js
```diff
@@ -1444,9 +1418,6 @@ class RenderWebGL extends EventEmitter {
         if ('skinId' in properties) {
             this.updateDrawableSkinId(drawableID, properties.skinId);
         }
-        if ('rotationCenter' in properties) {
```
fsih, line 1459: We might want to document the properties, but not necessarily in this PR

fsih: I guess if this function is going away it doesn't matter

apple502j:
FYI test is falling because playwright only supports node 10+ but it's still node 8

PR #542, opened 2020-01-09
Update EffectTransform.transformColor to match GPU "color" + "brightness" effects

adroitwhiz:
### Resolves

Resolves #578

### Proposed Changes

This PR changes `EffectTransform.transformColor` to match what the `sprite.frag` shader does with regards to the brightness and color effects.

In particular, it implements the "color" effect as a hue shift in HSV space, and the "brightness" effect as an addition in RGB space.

### Reason for Changes

`EffectTransform` is used in the CPU "touching color" pipeline, and so it should give the exact same results as the GPU.

### Test Coverage

CPU/GPU parity tests are blocked on #537

PR #582, opened 2020-04-09
Revert "Revert "Revert "Draw pen lines via fragment shader"""

fsih:
Reverts LLK/scratch-render#580

This will cause some compatibility issues with projects drawn with the pre-existing Scratch 3.0 jagged, oval pen, and Scratch team members want to spend some more time discussing the implications

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -44,6 +44,11 @@ const __projectionMatrix = twgl.m4.identity();
  */
 const __modelTranslationMatrix = twgl.m4.identity();
 
+/**
+ * Reused memory location for rotation matrix for building a model matrix.
+ * @type {FloatArray}
+ */
+const __modelRotationMatrix = twgl.m4.identity();
 
 /**
  * Reused memory location for scaling matrix for building a model matrix.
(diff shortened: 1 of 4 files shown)


adroitwhiz:
It's frustrating to wait a year for my PR to be merged, and then reverted because in the year it took for it to be merged, some people somewhere may have made projects relying on the behavior that existed because it wasn't merged :confused: 

This also raises a lot of questions about other behavior-effecting PRs like #479 and #542 -- those certainly change the way "touching" blocks behave. Will all future versions of Scratch, in whatever language they're written in, have to implement two different code paths for "touching" blocks depending on [the number of pixels in the thing being touched](https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L25), because of compatibility concerns with projects that rely on the buggy behavior of "touching" blocks?

It seems like the longer these bugs go without being fixed, or the more often that fixes which solve many problems are reverted because of a [single regression that may not have even been reproduced](https://github.com/LLK/scratch-render/issues/432), the greater the compatibility concern becomes and the greater the burden on future maintainers to keep "bug-for-bug compatibility" with the existing codebase.

This means that adding hacky user-facing changes is not the only way to increase tech debt-- simply leaving those changes in the codebase increases tech debt with each passing day as Scratchers come to rely on broken behavior.

fsih:
I totally agree! "Because the pen had a bug and spent some time not being a circle" definitely isn't a good reason to leave it as not a circle forever!

The way our release process is set up, most of our engineering team doesn't get to see the bug fixes they didn't work on until soon before we're supposed to release. If they have any questions or concerns about any of the bug fixes, we have to revert them to remove them from the release, because there aren't more than ~15 minutes to discuss each bug before release time. Reverting gives us a week to settle discussions before putting changes back in. It definitely doesn't mean we've decided not to put the change in anymore. I know it must be really frustrating for you!

Overall, the team is excited for this change, and everyone thinks the pen looks a lot better now! We haven't found any projects that would be broken by the pen changes. But there are a lot of pen projects, and it really sucks when your project breaks one day for no apparent reason, so we wanted extra time to really thoroughly test this change.

PR #588, opened 2020-04-24
Call draw() after resizing canvas

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/5678

### Proposed Changes

This PR calls `RenderWebGL.draw` in `RenderWebGL.resize` after resizing the canvas.

### Reason for Changes

When the color picker is active, [`stage.jsx`'s `componentDidUpdate` is called once per frame, which itself calls `RenderWebGL.resize`](https://github.com/LLK/scratch-gui/blob/develop/src/containers/stage.jsx#L97). This clears the canvas once per frame, causing rapid flickering. To avoid this, always redraw the canvas after resizing it.

An alternative would be to [set `preserveDrawingBuffer: true` in the context creation attributes](https://github.com/LLK/scratch-render/blob/develop/src/RenderWebGL.js#L108) but this seems to better solve the root issue (a one-frame delay between a canvas resize and a redraw).

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -223,6 +223,8 @@ class RenderWebGL extends EventEmitter {
         const pixelRatio = window.devicePixelRatio || 1;
         this._gl.canvas.width = pixelsWide * pixelRatio;
         this._gl.canvas.height = pixelsTall * pixelRatio;
+        // Resizing the canvas can cause it to be cleared, so redraw it.
+        this.draw();
     }
 
     /**
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -220,11 +220,20 @@ class RenderWebGL extends EventEmitter {
      * @param {int} pixelsTall The desired height in device-independent pixels.
      */
     resize (pixelsWide, pixelsTall) {
+        const {canvas} = this._gl;
         const pixelRatio = window.devicePixelRatio || 1;
-        this._gl.canvas.width = pixelsWide * pixelRatio;
-        this._gl.canvas.height = pixelsTall * pixelRatio;
-        // Resizing the canvas can cause it to be cleared, so redraw it.
-        this.draw();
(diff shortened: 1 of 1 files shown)


src/RenderWebGL.js
```diff
@@ -223,6 +223,8 @@ class RenderWebGL extends EventEmitter {
         const pixelRatio = window.devicePixelRatio || 1;
         this._gl.canvas.width = pixelsWide * pixelRatio;
         this._gl.canvas.height = pixelsTall * pixelRatio;
+        // Resizing the canvas can cause it to be cleared, so redraw it.
+        this.draw();
```
cwillisf, line 227: I like this idea! It seems unnecessary to resize & redraw every frame when the color picker is active, though. Are you aware of a reason we need to do that? If not, what do you think about something like this?
```suggestion
        const newWidth = pixelsWide * pixelRatio;
        const newHeight = pixelsTall * pixelRatio;
        if (this._gl.canvas.width != newWidth || this._gl.canvas.height != newHeight) {
            this._gl.canvas.width = pixelsWide * pixelRatio;
            this._gl.canvas.height = pixelsTall * pixelRatio;
            // Resizing the canvas can cause it to be cleared, so redraw it.
            this.draw();
        }
```

adroitwhiz: I was hesitant to include that in this PR in order to keep changes minimal considering it's a patch for a critical bug, but that's a good idea.

cwillisf:
FYI @adroitwhiz: since it's Friday afternoon/evening right now, most of the Scratch Team isn't available. It might be Monday before we really move on this but in my opinion it might be a candidate for hotfixing early in the week. We'll see what others on the team say...

src/RenderWebGL.js
cwillisf: Ah, good point. I think this is OK but I admire your restraint! :D

PR #590, opened 2020-04-25
Tweak pen shader to avoid float precision issues

### LINKED ISSUE #589: Pen dots in the lower right quadrant are not drawn on iOS Safari
### Expected Behavior

Pen dots should always be drawn

### Actual Behavior

On iOS Safari, pen dots are not drawn in the lower right quadrant-- pen lines render fine:
293F441E-3339-46D0-B4CD-191C97E4AD17
(Screenshot taken from [this project](https://scratch.mit.edu/projects/388753987/))

### Steps to Reproduce

1. Create a project that draws pen dots all around the stage (with "pen down" immediately followed by "pen up")-- see [this project](https://scratch.mit.edu/projects/96967170/) as an example
2. Observe that on iOS Safari, the dots are not drawn in the lower right quadrant

iOS Safari

### LINKED ISSUE #595: Pen lines over ~100 pixels long will not render on (some?) Android devices
### Expected Behavior

All pen lines should show up.

### Actual Behavior

On (certain?) Android devices, lines over ~100 pixels long will not be drawn:
image

### Steps to Reproduce

1. Create a project
2. Add the pen extension
3. Click "pen down"
4. Make the active sprite move >100ish steps
5. Observe that only a pen dot is drawn, whereas both a pen dot and a pen line should be drawn

I've also created [a repro project that tests lines of different lengths](https://scratch.mit.edu/projects/392244913/).

Android, Chrome + Firefox + others?

Tested on a OnePlus 6T with an Adreno 630 GPU-- the specific model of GPU may play a role in this as well.

adroitwhiz:
### Resolves

Resolves #589
Resolves #595

### Proposed Changes

This PR tweaks the "epsilon" logic used in the shaders to avoid numerical instability and division by zero on low-precision hardware.

### Reason for Changes

~~If you revert #586, the commit message will be dangerously long~~

iOS Safari treats "medium precision" as much lower precision than other browsers, causing some of the pen line formulas to bug out and pen dots to not be drawn. This PR tweaks those formulas to avoid such numerical issues.

As a nice side effect, pen dots' quads will now be rectangular rather than diamond-shaped, which may provide a very slight speed boost.

There's another issue on some??? GPUs where certain calculations in the line shader could overflow the floating-point range if the lines are long enough. I confirmed this on a OnePlus 6T with an Adreno 630 GPU, and someone else reported this on an ARM Chromebook on the Bugs and Glitches forum; not sure where else it may occur.

### Test Coverage

Right now there's no way to test shaders across different hardware. It might be a good idea to look into whether that can be done.

I've manually tested this on an iPhone that was buggy before, and it fixes the issue.

### DIFF
--- a/src/shaders/sprite.vert
+++ b/src/shaders/sprite.vert
@@ -42,7 +42,10 @@ void main() {
 	// Rotate quad to line angle
 	vec2 pointDiff = u_penPoints.zw - u_penPoints.xy;
 	// Ensure line has a nonzero length so it's rendered properly
+	// As long as either component is nonzero, the line length will be nonzero
 	pointDiff.x = abs(pointDiff.x) < epsilon ? epsilon : pointDiff.x;
+	// The `normalized` vector holds rotational values equivalent to sine/cosine
+	// We're applying the standard rotation matrix formula to the position to rotate the quad to the line angle
 	vec2 normalized = pointDiff / max(lineLength, epsilon);
 	position = mat2(normalized.x, normalized.y, -normalized.y, normalized.x) * position;
 	// Translate quad
--- a/src/shaders/sprite.vert
+++ b/src/shaders/sprite.vert
@@ -40,7 +40,10 @@ void main() {
 	position -= expandedRadius;
 
 	// Rotate quad to line angle
-	vec2 normalized = (u_penPoints.zw - u_penPoints.xy + epsilon) / (lineLength + epsilon);
+	vec2 pointDiff = u_penPoints.zw - u_penPoints.xy;
+	// Ensure line has a nonzero length so it's rendered properly
+	pointDiff.x = abs(pointDiff.x) < epsilon ? epsilon : pointDiff.x;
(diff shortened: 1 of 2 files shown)


src/shaders/sprite.vert
```diff
@@ -40,7 +40,10 @@ void main() {
 	position -= expandedRadius;
 
 	// Rotate quad to line angle
-	vec2 normalized = (u_penPoints.zw - u_penPoints.xy + epsilon) / (lineLength + epsilon);
+	vec2 pointDiff = u_penPoints.zw - u_penPoints.xy;
+	// Ensure line has a nonzero length so it's rendered properly
+	pointDiff.x = abs(pointDiff.x) < epsilon ? epsilon : pointDiff.x;
+	vec2 normalized = pointDiff / max(lineLength, epsilon);
```
fsih, line 49: Add epsilon to line length rather than changing pointdiff.x for readability

PR #594, opened 2020-05-05
Handle subpixel viewboxes for SVG skins

### LINKED ISSUE #143: Render issue with Scratch Cat's paw?
### Expected Behavior

_Please describe what should happen_

Default sprite in scratch 2:
image

### Actual Behavior

_Describe what actually happens_
In scratch 3:
image

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Open up https://llk.github.io/scratch-gui/

_e.g. Mac OS 10.11.6 Safari 10.0_
Chrome 58.0.3029.110, Mac 10.12.4 

I tried this on Firefox and couldn't reproduce, so it might be something specific about chrome webgl?

### LINKED ISSUE #233: Some parts of Scratch Cat's foot is not drawn
### Expected Behavior

This is the cat in Paint Editor. It's OK.
Paint Editor is OK
### Actual Behavior

This is the cat in Stage, and wrong.
Stage is wrong
### Steps to Reproduce
Open the page.
Firefox 57.0.4 windows7 HomePremium SP1 64bit

### LINKED ISSUE #427: Rounding of vector bounding box dimensions causes jittering of sprite

### LINKED ISSUE #533: Use js function log2
Is this equivalent to taking the rounded log base 2? https://github.com/LLK/scratch-render/blob/b1274d5820f00a198684f9c592f7db95c36e4368/src/SVGSkin.js#L117

I think it would be more straightforward to use log2

### LINKED ISSUE #639: Costumes are cut off at the bottom and right on the stage
### Expected Behavior
Whole costume should show

### Actual Behavior
A couple pixels get cut off on the bottom and right
Screen Shot 2020-06-11 at 12 05 35

### Steps to Reproduce
Draw a costume, and look carefully at the bottom and right edges

all

adroitwhiz:
### Resolves

Resolves #143
Resolves #233
Resolves #427
Resolves #533
Resolves #639

### Proposed Changes

- ~~Clean up the SVGSkin code a bit~~ Done in #649
- Set the minimum SVG texture scale to 1/8 its normal size
  - This is mentioned in #550--the "native size" of the rendered sprite must be a multiple of 1/the smallest scale, so this means that all SVG skins' dimensions will be rounded up to a multiple of 8. Setting this lower (to, for instance, 1/32 scale) would require all SVG skins' dimensions to be rounded up to a multiple of 32. I think 1/8 is a good tradeoff between skins not looking pixelated at small sizes and minimizing the amount of "slack space" around the edges.
- Use the adjusted bounds + center from the algorithm in #550 on SVG skins.
- Pass "logical bounds" into the shader so it can calculate "logical coordinates" under which distortion effects are applied.
  - SVG skins' bounds are now expanded, but we want things like the whirl and fisheye effects' center to still be positioned in the middle of the original, non-expanded bounds. We also don't want the pixelate or mosaic effects to go beyond the non-expanded bounds. To do this, we tell the shader what the original bounds are.
- Separate skins' size and rotation center into a "native size" + "native rotation center" and "quad size" + "quad rotation center'.
  - The "native size" and "native rotation center" refer to the un-expanded bounds. They are used to calculate skins' AABBs and when the VM wants to get a skin's size. They do not depend on details like the amount of "slack space" present around the adjusted texture, and as such are used for fencing sprites.
  - The "quad size" and "quad rotation center" refer to the expanded bounds. They are used for everything else, and are so named because they refer to the size of the actual rendered quadrilateral. I'm not sure what else to call them-- "texture size" seems misleading. Any ideas?

### Reason for Changes

As I talked about in #550, these changes seem to be the best way to fix the jittering problem that vector costumes have at the moment. These do add some complexity but I'm not sure how to avoid it, so I could use your help in that area.

### Test Coverage

Tests ~~are blocked on #537 and~~ still need some discussion on what should be tested


src/SVGSkin.js
```diff
@@ -159,6 +159,14 @@ class SVGSkin extends Skin {
             this._rotationCenter[0] = adjustedCenter[0];
             this._rotationCenter[1] = adjustedCenter[1];
 
+            // Set "logical bounds" as fractions of the adjusted bounding box size
```
fsih: Add more commentary here

src/shaders/sprite.frag
```diff
@@ -154,8 +164,16 @@ void main()
 	}
 	#endif // ENABLE_fisheye
 
+	// After doing all distortions in "logical texture space", convert back to actual texture space
+	texcoord0 = (texcoord0 * (u_logicalBounds.zw - u_logicalBounds.xy)) + u_logicalBounds.xy;
+
 	gl_FragColor = texture2D(u_skin, texcoord0);
 
+	#ifdef ENABLE_pixelate
+	// Ensure that the pixels don't extend outside the "logical bounds"
```
fsih, line 173: Add comment about why we aren't worried about <0

src/shaders/sprite.vert
```diff
@@ -63,5 +65,6 @@ void main() {
 	#else
 	gl_Position = u_projectionMatrix * u_modelMatrix * vec4(a_position, 0, 1);
 	v_texCoord = a_texCoord;
+	v_logicalCoord = (a_texCoord - u_logicalBounds.xy) / (u_logicalBounds.zw - u_logicalBounds.xy);
```
fsih: add comment here

src/Skin.js
```diff
@@ -82,26 +82,54 @@ class Skin extends EventEmitter {
...
+        // TODO: is it significantly expensive to go through two nested getters for every non-vector sprite?
+        return this.nativeSize;
+    }
+
+    /**
+     * @returns {boolean} True if this skin's "quadrilateral size" is always the same as its "native size",
+     * false if it differs.
+     * Vector-based skins like the SVGSkin and TextBubbleSkin will have different "quadrilateral" and "native" sizes.
+     */
+    get quadSizeIsNativeSize () {
```
fsih: Remove?

src/Drawable.js
```diff
@@ -564,12 +568,64 @@ class Drawable {
...
+                rightEdge + x,
+                bottomEdge + y,
+                topEdge + y
+            );
+
+            this._aabbDirty = false;
+        }
+
+
+        result = result || this._aabb;
```
fsih: Always instantiate new instead of passing this._aabb

src/Drawable.js
```diff
@@ -102,6 +102,9 @@ class Drawable {
         this._inverseTransformDirty = true;
         this._visible = true;
 
+        this._aabbDirty = true;
```
fsih, line 103: transformDirty means that one or more of rotationTransformDirty, skinScaleDirty, rotationCenterDirty.
transformDirty implies aabbDirty, but aabbDirty doesn't imply transformDirty

src/Drawable.js
```diff
@@ -564,12 +568,64 @@ class Drawable {
...
+            const sin = this._rotationMatrix[1];
+
+            // TODO: this can likely be golfed
+            const origLeft = center[0] * this._scale[0] / 100;
+            const origTop = -center[1] * this._scale[1] / 100;
+            const origRight = (center[0] - size[0]) * this._scale[0] / 100;
+            const origBottom = (size[1] - center[1]) * this._scale[1] / 100;
+
+            const leftCos = origLeft * cos;
+            const rightCos = origRight * cos;
```
fsih: It bothers me that top and bottom are grouped but left and right are intermingled in the ordering

src/Drawable.js
```diff
@@ -564,12 +568,64 @@ class Drawable {
...
+
+            // Calculate the rotated "logical" bounding box
+            const size = this.skin.nativeSize;
+            const center = this.skin.nativeRotationCenter;
+
+            // TODO: This is a tad bit janky but saves some math. Worth it?
+            const cos = this._rotationMatrix[0];
+            const sin = this._rotationMatrix[1];
+
+            // TODO: this can likely be golfed
```
fsih: Comment

src/SVGSkin.js
```diff
@@ -156,13 +156,17 @@ class SVGSkin extends Skin {
...
-                return;
-            }
+        this._svgRenderer.loadString(svgData, false /* fromVersion2 */);
+        const svgSize = this._svgRenderer.size;
+        if (svgSize[0] === 0 || svgSize[1] === 0) {
+            super.setEmptyImageData();
+            return;
+        }
+
+        if (typeof rotationCenter === 'undefined') rotationCenter = this.calculateRotationCenter();
```
fsih, line 170: calculateRotationCenter () {
        return [this.size[0] / 2, this.size[1] / 2];
    }

src/EffectTransform.js
```diff
@@ -224,8 +224,8 @@ class EffectTransform {
         }
         if ((effects & ShaderManager.EFFECT_INFO.pixelate.mask) !== 0) {
             // vec2 pixelTexelSize = u_skinSize / u_pixelate;
-            const texelX = skinUniforms.u_skinSize[0] * uniforms.u_pixelate;
```
fsih: Remember to put this separate PR in at some point

src/EffectTransform.js
```diff
@@ -217,12 +217,15 @@ class EffectTransform {
         dst[1] = (dst[1] - skinUniforms.u_logicalBounds[1]) /
             (skinUniforms.u_logicalBounds[3] - skinUniforms.u_logicalBounds[1]);
 
-        if ((effects & ShaderManager.EFFECT_INFO.mosaic.mask) !== 0) {
+        const pointInsideLogicalBounds = dst[0] >= 0 && dst[0] <= 1 && dst[1] >= 0 && dst[1] <= 1;
```
cwillisf: Would it make sense to have the caller do the remap to logical coordinates and then this check, and not even call `transformPoint` if the logical coordinate is outside the bounds? I'm worried about side effects of discontinuous transformations in (for example) convex hull calculation.

adroitwhiz: Done in https://github.com/LLK/scratch-render/pull/594/commits/036895581e72045b95957e656eec64f117e8eca5

adroitwhiz:
@fsih @cwillisf This should be ready now that scratch-svg-renderer is all split up and we can do the viewbox logic here! I've also tried my best to address the review comments. Also note that this depends on #745.

I remember you not liking the names `quadSize` and `quadRotationCenter`--any ideas what I could use instead? `textureSize`/`textureRotationCenter` seems misleading because that seems to refer to the texture's actual dimensions, which isn't what `quadSize` and `rotationCenter` refer to.

Issue #577, opened 2020-03-22
Text bubbles are counted for "touching color" purposes in 3.0, but not in 2.0


adroitwhiz:
In Scratch 2.0, text bubbles don't count for the purposes of "touching color" blocks, e.g. if a sprite is touching a text bubble, that doesn't count as touching white or grey or any other color on that text bubble; they are ignored completely.

In Scratch 3.0, however, "touching color" blocks can return true if the sprite is touching a text bubble's color.

This raises the question: should this be noted as a compatibility change between 2.0 and 3.0 and the current behavior kept, or should the behavior be changed to match 2.0?

<details>
<summary>
A unit test was supposed to verify that this is the case in 3.0 as well, but...
</summary>

Due to a series of weird coincidences and oversights, this unit test has erroneously passed when it should not have, and text bubbles have been part of "touching color" tests for, as far as I can tell, the entire 3.0 lifespan.
- The unit test runs on the very first frame. Unfortunately, the SVG-based text bubbles implemented when the test was first created had a delay of about 1-2 frames before showing up.
- The canvas-based text bubbles introduced in #451 are created synchronously, and so they should have revealed the error, however...
- Due to the same issue as in #544, text bubble silhouettes are not created until the bubbles have been rendered. This wouldn't be a problem, since this bug only manifests on the CPU path, but...
- Since the CPU/GPU parity testing in #419 was reverted before #544 was merged, the test never runs on the GPU path.

</details>

PR #598, opened 2020-05-07
Run integration tests on both the CPU and GPU (again)

### LINKED ISSUE #587: Run tests on both the CPU and GPU again
Unit tests should be run on both the CPU and GPU implementations of the "touching" functions to ensure that they both function properly. This was done in #419, but was later reverted.

This is blocked on:
- #576 - without this merged, the touching color test will fail on the CPU.
- #577 - there's currently an "is not touching text bubble" test that will fail on the GPU. Text bubble behavior needs to be ironed out.
- #537 - this isn't necessarily a blocker, but it may be easier to merge #537 and then reimplement CPU/GPU testing than to reapply https://github.com/LLK/scratch-render/pull/419/commits/f9428ee096b7d85f0fe0dc35d6e9ace1c2db59cc and then re-Puppeteer-ify it.

adroitwhiz:
### Resolves

Resolves #587

### Proposed Changes

This PR adapts the changes in https://github.com/LLK/scratch-render/pull/419/commits/f9428ee096b7d85f0fe0dc35d6e9ace1c2db59cc to the new Playwright-based testing code.

### Reason for Changes

This should increase test coverage and hopefully prevent future bugs that only occur on either the CPU or GPU and as such were not caught by the previous testing behavior.

### Test Coverage

Tests currently fail, which is good because they're catching existing bugs. #576 will need to be merged before tests will pass again. For now, I've disabled the "doesn't touch text bubble" test because there are potential compatibility concerns I want to work out before fixing that (https://github.com/LLK/scratch-render/issues/577).


test/integration/scratch-tests.js
```diff
@@ -4,55 +4,13 @@ const test = require('tap').test;
...
-                    await new Promise(resolve => setTimeout(resolve, 50));
-                }
-
-                return messages;
-            });
-    });
-
-    // Map string messages to tap reporting methods. This will be used
+const checkOneGpuMode = (t, says) => {
+// Map string messages to tap reporting methods. This will be used
```
fsih: Indentation?

cwillisf (approved):
👍 approved pending successful automated tests on CI

PR #597, opened 2020-05-07
Switch from Chromeless to Playwright for tests

adroitwhiz:
### Resolves

Resolves #564
Closes #537

### Proposed Changes

Like #537, but with Playwright instead of Puppeteer. The JavaScript ecosystem is totally normal. There is nothing wrong with the JavaScript ecosystem.

### Reason for Changes

See #537-- Chromeless was deprecated mid-2018 and has not been updated since then.

### Test Coverage

Yup, that's what it does.

PR #603, opened 2020-05-08
Revert "Switch from Chromeless to Playwright for tests"

adroitwhiz:
Reverts LLK/scratch-render#597

Weird Travis caching stuff to be investigated another time.


cwillisf (approved):
👍
Thanks for being on top of this and noticing the caching trouble right away! I look forward to getting this back in once we figure that part out...

PR #621, opened 2020-05-16
Take render target dimensions into account when drawing

### LINKED ISSUE #545: drawableScale should consider current render target

### LINKED ISSUE #558: Stamp Quality Differs Depending on Stage Size
### Expected Behavior

Stamp quality should not noticeably differ based on the resolution or size of the stage (such that using full screen mode increases the stage size/resolution).

### Actual Behavior

Stamp quality appears to significantly improve in full screen mode compared to normal screen and small screen mode.

### Steps to Reproduce

Go to https://llk.github.io/scratch-gui/develop/#377548941. Pressing 1 will draw on the left side, 2 in the middle, and 3 on the right side. Pressing s will change the text size. Clicking the green flag will erase everything. To see the issue, press 1 while in small screen mode in the editor, press 2 while in normal screen mode, and press 3 while in full screen mode. The differences should be obvious while in full screen mode, but still can be seen even in small screen mode with a 1920x1080 monitor.

This is the result on my system when in fullscreen:
stamp bug

Note that this issue occurs even when not using rotation. Text is also not required, since the costumes are SVG paths rather than text.

Old project: https://llk.github.io/scratch-gui/develop/#371843546

Windows 10 Home 1903
Chrome 80.0.3987.122
1920x1080 resolution with 100% scaling

adroitwhiz:
### Resolves

Resolves #545
Resolves #558

### Proposed Changes

This PR changes `_drawThese` to accept optional `framebufferWidth` and `framebufferHeight` parameters, and default to the "native size" (480x360) if they're not given.

These dimensions are used to calculate drawables' "framebuffer-space" scales.

### Reason for Changes

See #545.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -653,7 +653,10 @@ class RenderWebGL extends EventEmitter {
         gl.clearColor.apply(gl, this._backgroundColor4f);
         gl.clear(gl.COLOR_BUFFER_BIT);
 
-        this._drawThese(this._drawList, ShaderManager.DRAW_MODE.default, this._projection);
+        this._drawThese(this._drawList, ShaderManager.DRAW_MODE.default, this._projection, {
+            framebufferWidth: gl.canvas.width,
+            framebufferHeight: gl.canvas.height
+        });
         if (this._snapshotCallbacks.length > 0) {
             const snapshot = gl.canvas.toDataURL();
             this._snapshotCallbacks.forEach(cb => cb(snapshot));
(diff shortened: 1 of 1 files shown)


adroitwhiz:
@cwillisf This resolves #558, which is priority 2, so I want to make sure this hasn't fallen off your radar (given it has no "needs review" tag and is sandwiched in between several Dependabot PRs).

cwillisf:
It looks to me like the `extractDrawable` methods should provide buffer dimensions when calling `_drawThese` since it might use the query buffer OR a new, potentially very large buffer.

adroitwhiz:
Good catch! I've done this for `extractDrawableScreenSpace`--does anything still use `extractDrawable` other than the playground? I can't see anything in any public repos. Regardless, it looks like `extractDrawable` does everything at the "Scratch-space" resolution anyway, so I think it should be left as is.

fsih (approved):
LG, thanks!

We're planning to not deploy changes during December, due to the annual hour of code and holidays, so will be holding off on merging until January

PR #624, opened 2020-05-19
Move `updateCPURenderAttributes` calls to where they're actually needed

### LINKED ISSUE #623: Current CPU render code relies on implicit behavior
### Expected Behavior

The `isTouchingDrawables` and `isTouchingColor` functions should update the target drawable's silhouette and inverse matrix, as they require them to be up-to-date in the first place.

### Actual Behavior

Those functions currently rely on the `_touchingBounds` function updating those properties on the target drawable. `_touchingBounds` is called by `_candidatesTouching` which in turn is called by `isTouchingDrawables` and `isTouchingColor`.

Since `_touchingBounds` doesn't require those properties to be up-to-date itself, it could be changed in the future to stop updating them, which would break "touching color" and "touching sprite" as they rely on it doing so.

adroitwhiz:
### Resolves

Resolves #623

### Proposed Changes

This PR moves `updateCPURenderAttributes` calls from the `_touchingBounds` function to `isTouchingColor` and `isTouchingDrawable`, and moves the `updateCPURenderAttributes` call in `_getConvexHullPointsForDrawable` to after the no-points-in-hull check.

### Reason for Changes

`_touchingBounds` doesn't actually require the passed drawable's matrix or silhouette to be up-to-date. However, `isTouchingColor` and `isTouchingDrawables` relied on it updating those. For clarity (and to prevent things from breaking if a future coder decided to "optimize" `_touchingBounds`), these calls have been moved (in the form of calls to the new `updateCPURenderAttributes` function) to the places that *actually* require the drawable's matrix and silhouette to be updated.

`_getConvexHullPointsForDrawable` doesn't do anything that requires the drawable's CPU render stuff to be up-to-date until after the no-points-in-hull check, so it's safe to not update them until after it.

### DIFF
--- a/src/RenderWebGL.js
+++ b/src/RenderWebGL.js
@@ -769,6 +769,8 @@ class RenderWebGL extends EventEmitter {
         const color = __touchingColor;
         const hasMask = Boolean(mask3b);
 
+        drawable.updateCPURenderAttributes();
+
         // Masked drawable ignores ghost effect
         const effectMask = ~ShaderManager.EFFECT_INFO.ghost.mask;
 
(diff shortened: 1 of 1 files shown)


fsih (approved):
LG, thanks!

PR #649, opened 2020-06-22
Always use SVG mipmaps that match or exceed the skins' screen-space size

### LINKED ISSUE #533: Use js function log2

### LINKED ISSUE #620: Should vector skins/costumes always be rendered with a resolution greater than their screen-space dimensions?
### Expected Behavior

Vector skins/costumes should appear sharp at all scales.

### Actual Behavior

Vector skins are internally represented by a series of textures at various resolutions. These textures are allowed to appear on-screen at up to 1.5x their resolution, meaning that at worst, 1 "texture pixel" will be magnified to 1.5 "screen pixels". This results in vector sprites often appearing blurry:
image

This blurriness is particularly bad for things like text, which results in issues like https://github.com/LLK/scratch-render/issues/558.

This could be changed so that the textures must always be at a higher resolution than, or the same resolution as, their "screen-space" dimensions. This would consume more RAM, but the improvement in visual fidelity may well be worth it.

Keep in mind that until last October, [vector costumes were rendered at least twice as large as they should have been on high-DPI devices like phones](https://github.com/LLK/scratch-svg-renderer/pull/84), so this shouldn't have an unprecedented RAM impact.

### Steps to Reproduce

1. Create a new project
2. With the default cat sprite selected, click on the "change size by 10" block in the Blocks Palette while observing the cat sprite
3. Notice that as you continue clicking the block, the cat gets blurrier as it grows, eventually becoming crisp again if you click the block enough

All

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-render/issues/533
Resolves https://github.com/LLK/scratch-render/issues/620

### Proposed Changes

This PR changes the SVG mipmap logic to always use a mipmap with a greater or equal resolution to that of the sprite as displayed on-screen.

### Reason for Changes

This makes vector sprites appear less blurry.

### Test Coverage

To be tested manually

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -119,7 +119,11 @@ class SVGSkin extends Skin {
         const scaleMax = scale ? Math.max(Math.abs(scale[0]), Math.abs(scale[1])) : 100;
         const requestedScale = Math.min(scaleMax / 100, this._maxTextureScale);
 
-        const mipLevel = Math.max(Math.round(Math.log2(requestedScale)) + INDEX_OFFSET, 0);
+        // Math.ceil(Math.log2(scale)) means we use the "1x" texture at (0.5, 1] scale,
+        // the "2x" texture at (1, 2] scale, the "4x" texture at (2, 4] scale, etc.
+        // This means that one texture pixel will always be between 0.5x and 1x the size of one rendered pixel,
+        // but never bigger than one rendered pixel--this prevents blurriness from blowing up the texture too much.
+        const mipLevel = Math.max(Math.ceil(Math.log2(requestedScale)) + INDEX_OFFSET, 0);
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
👍 LGTM! Thanks for making the math easier to read, too :)

fsih (approved):
Looks great! LG from both me and Chris if it passes performance testing ok

fsih:
branch: https://adroitwhiz.github.io/scratch-gui/no-vector-blur/

PR #654, opened 2020-07-06
Don't include text bubbles in touching queries

### LINKED ISSUE #577: Text bubbles are counted for "touching color" purposes in 3.0, but not in 2.0

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-render/issues/577

### Proposed Changes

This changes `_candidatesTouching` to exclude text bubbles, and fixes text bubbles to properly update their silhouettes.

### Reason for Changes

* Change `_candidatesTouching` to exclude text bubbles
  * This changes the behavior of all "touching" blocks to not count text bubbles.
* Add `updateSilhouette` to `TextBubbleSkin`
  * This is why the test *wasn't* failing when it should have. It was checking if it was touching the text bubble on the very first frame, before the text bubble was even rendered. At that point, the text bubble's silhouette didn't exist, as it was only generated when the text bubble was rendered. Now, the text bubble's silhouette is generated whenever `updateSilhouette` is called.

### Test Coverage

Re-enabled the "text bubbles are not touched" test


src/TextBubbleSkin.js
```diff
@@ -241,6 +241,11 @@ class TextBubbleSkin extends Skin {
         this._renderedScale = scale;
     }
 
+    updateSilhouette (scale = [100, 100]) {
```
fsih, line 244: Why was this needed?

adroitwhiz: This ensures that text bubbles' "silhouettes" behave consistently on the CPU render path. Since text bubbles are no longer rendered on the CPU path, this is technically not necessary but should prevent future weirdness like what happened with the test here where text bubbles appeared to not count as being touched for the first frame.

fsih (approved):
LG, thanks!

PR #667, opened 2020-07-23
Move pen point difference calculation from shader code to JS

### LINKED ISSUE #642: Some pen strokes result in large, blurred circles on iPhone Safari
### Expected Behavior

Pen strokes should not result in large, blurred circles.

Example:
IMG_0032

### Actual Behavior

Pen strokes often result in large, blurred circles.

Example:
IMG_1562

### Steps to Reproduce

On an iPhone, run one of the following projects on the project page:
- https://scratch.mit.edu/projects/278999727/
- https://scratch.mit.edu/projects/276932192/
- https://scratch.mit.edu/projects/318234147/
- https://scratch.mit.edu/projects/262564340/
- https://scratch.mit.edu/projects/236128688/

This doesn’t seem to be an issue limited to a few specific projects; most projects using the pen should produce this issue.

iPhone XS iOS 13.5.1 Safari

Cannot reproduce on a 2019 iPad on iOS 13.5.1

adroitwhiz:
### Resolves

Resolves #642

### Proposed Changes

This changes the pen line code so that the difference between pen points is calculated in JS instead of in the shader code.

I also stopped passing `u_penPoints` into the fragment shader (it hasn't needed that since last time I fixed this; I just forgot to remove it then), and clarified some of the comments.

### Reason for Changes

If the pen points' coordinates are both somewhat large, calculating the difference between them will be inaccurate if done using the lower floating-point precision on GPUs. This messes up vertex shader calculations further down the line.

I'm pleasantly surprised by how simple the fix was here--it shouldn't decrease performance at all.

### Test Coverage

Tested manually on an iPad Mini 5 (A2133) that was previously exhibiting the issue.

### DIFF
--- a/src/PenSkin.js
+++ b/src/PenSkin.js
@@ -323,7 +323,7 @@ class PenSkin extends Skin {
             u_lineColor: __premultipliedColor,
             u_lineThickness: penAttributes.diameter || DefaultPenAttributes.diameter,
             u_lineLength: lineLength,
-            u_penPoints: [x0, -y0, x1, -y1],
+            u_penPoints: [x0, -y0, lineDiffX, -lineDiffY],
             u_stageSize: this.size
         };
 
--- a/src/shaders/sprite.frag
+++ b/src/shaders/sprite.frag
@@ -37,7 +37,6 @@ uniform float u_ghost;
 uniform vec4 u_lineColor;
 uniform float u_lineThickness;
 uniform float u_lineLength;
-uniform vec4 u_penPoints;
 #endif // DRAW_MODE_line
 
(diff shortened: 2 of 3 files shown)


adroitwhiz:
After some investigation, it seems like the root cause of this is...

*sigh...*

Affected devices think that if you subtract a number from itself, you don't get 0.
This seems to be a shader miscompilation on Apple's side. This PR works around it. I'll try to contact Apple about this bug.

fsih (approved):
Nice! Looks reasonable to me!

cwillisf (approved):
LGTM!

PR #676, opened 2020-08-11
Put back in "Merge pull request #489 from adroitwhiz/touching-white-fixes"

### LINKED ISSUE #680: scratch-render tests fail locally on node 12
### Expected Behavior
Tests should pass regardless of node version

### Actual Behavior
Tests fail, something about requizzle

### Steps to Reproduce
`npm run test` in scratch-render
Screen Shot 2020-08-20 at 15 43 17

Mac OSX

fsih:
Reverts LLK/scratch-render#674

Fixes #680


src/RenderWebGL.js
```diff
@@ -745,12 +766,20 @@ class RenderWebGL extends EventEmitter {
...
     isTouchingColor (drawableID, color3b, mask3b) {
         const candidates = this._candidatesTouching(drawableID, this._visibleDrawList);
-        if (candidates.length === 0) {
+
+        let bounds;
+        if (colorMatches(color3b, this._backgroundColor3b, 0)) {
+            // If the color we're checking for is the background color, don't confine the check to
+            // candidate drawables' bounds--since the background spans the entire stage, we must check
+            // everything that lies inside the drawable.
+            bounds = this._touchingBounds(drawableID);
```
fsih, line 775: The issue is here, touchingBounds can return null: https://github.com/LLK/scratch-render/blob/954cfff02b08069a082cbedd415c1fecd9b1e4fb/src/RenderWebGL.js#L1385

Then we reference bounds.width on line 792

fsih:
I am trying to figure out why row 1 and column 1 of the background silhouette are full of transparent pixels (which are getting detected as white) and it's the rabbit hole

fsih:
Rabbit holes are actually not very deep or complicated? https://i.pinimg.com/originals/d9/cc/2a/d9cc2ab7fff45888ef3fef5fc5e92ed1.jpg

fsih:
I think the first row and column being detected as transparent is coming from floating point calculation issues. Here in sampleColor4b the localPosition is coming out to barely < 0, so it's returning transparent.
Screen Shot 2020-08-12 at 20 39 26
Screen Shot 2020-08-13 at 00 29 05

Here's what's in the matrix.
Screen Shot 2020-08-13 at 00 31 13

Edit: I was having a lot of frustration testing using Blue Sky 2, but I realized it actually is offset from the left. It has transparent padding all around and a viewbox that starts at x < 0, y < 0

Edit: filed https://github.com/LLK/scratch-render/issues/688 for this

src/Drawable.js
```diff
@@ -37,6 +37,8 @@ const getLocalPosition = (drawable, vec) => {
     // localPosition matches that transformation.
     localPosition[0] = 0.5 - (((v0 * m[0]) + (v1 * m[4]) + m[12]) / d);
     localPosition[1] = (((v0 * m[1]) + (v1 * m[5]) + m[13]) / d) + 0.5;
+    if (Math.abs(localPosition[0] < 1e-8)) localPosition[0] = 0;
+    if (Math.abs(localPosition[1] < 1e-8)) localPosition[1] = 0;
```
adroitwhiz, line 43: I think this can be removed once #479 goes in--maybe add a TODO?

fsih:
"Touches the colorfully bent semi-transparent crab" is failing

cwillisf:
I was comparing some of this code to #479 and I think the best approach might be to first merge that PR then look at this again. I think the way that #479 does the texture clamping in the `Silhouette.js` `get*` functions is probably better than the approach I suggested here, where you're clamping in `Drawable#sampleColor4b`. In fact, I now feel like all the `if (localPosition...` code in `sampleColor4b` -- even the old code from before this PR -- should just be removed in favor of the stuff in `Silhouette.js`. There might be some subtlety that I'm missing though... hopefully the colorfully bent semi-transparent crab will know for sure.

fsih:
The colorfully bent semi-transparent crab knows best

cwillisf:
Hmm, another idea that might be worth trying: what happens if you just remove the clamping in `Drawable#sampleColor4b` without #479's changes? We won't have all the other fixes but that might be sufficient to unblock this PR...

fsih:
The colorfully bent semi-transparent crab has spoken and it is Not Ok

src/Drawable.js
```diff
@@ -720,14 +720,6 @@ class Drawable {
...
     static sampleColor4b (vec, drawable, dst, effectMask) {
         const localPosition = getLocalPosition(drawable, vec);
-        if (localPosition[0] < 0 || localPosition[1] < 0 ||
-            localPosition[0] > 1 || localPosition[1] > 1) {
-            dst[0] = 0;
-            dst[1] = 0;
-            dst[2] = 0;
-            dst[3] = 0;
-            return dst;
-        }
```
cwillisf: So... it turns out I steered you wrong when I recommended removing this. I was thinking about texture coordinates being out of range (which should be and is handled at the Silhouette layer) when I should have been thinking about geometric coordinates being out of range (which is what this `if` handles). I think bringing this back in will fix the crab test.

package.json
```diff
@@ -34,7 +34,7 @@
     "eslint": "^4.6.1",
     "eslint-config-scratch": "^5.0.0",
     "gh-pages": "^1.0.0",
-    "jsdoc": "^3.5.5",
+    "jsdoc": "^3.6.0",
```
cwillisf, line 37: Might want to add "resolves #680" to the PR description... at least I think that's what this is...?

cwillisf (approved):
Looks good! I'm glad the crab is pleased 🦀

package.json
fsih: I broke this out into https://github.com/LLK/scratch-render/pull/683

PR #723, opened 2020-11-09
Update eslint, babel-eslint, and eslint-scratch-config and fix resulting errors

seotts:
The fixes obey the rules `no-prototype-builtins`, `prefer-spread`, use arrow functions when required and remove build-in global variable declaration in comments, because the new version doesn't require that.

These changes are primarily in tests, `RenderWebGL.js` and the playground.


cwillisf (approved):
👍 thanks!

I did notice a slight change in the spacing of the "You clicked on" log message but I don't think that's a big deal. If anybody's trying to parse exact `console.log` messages from the playground then they get what they deserve ;)

src/playground/playground.js
```diff
@@ -184,7 +184,7 @@ canvas.addEventListener('mousemove', event => {
 canvas.addEventListener('click', event => {
     const mousePos = getMousePosition(event, canvas);
     const pickID = renderer.pick(mousePos.x, mousePos.y);
-    console.log('You clicked on ' + (pickID < 0 ? 'nothing' : 'ID# ' + pickID));
+    console.log(`You clicked on ${(pickID < 0 ? 'nothing' : `ID#${pickID}`)}`);
```
seotts:
```suggestion
    console.log(`You clicked on ${(pickID < 0 ? 'nothing' : `ID# ${pickID}`)}`);
```

PR #745, opened 2021-01-08
Move SVG renderer logic back into SVGSkin

### LINKED ISSUE #550: Subpixel SVG costume handling: a writeup

adroitwhiz:
### Resolves

A step towards #550

### Proposed Changes

This PR moves the logic for actually rendering SVGs back into `SVGSkin` from `SvgRenderer`, and removes the dependency on the `SvgRenderer` class.

### Reason for Changes

This will allow for future modifications to the way SVGs are drawn, which are necessary to fix #550.

### Test Coverage

Tested manually

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -197,6 +197,7 @@ class SVGSkin extends Skin {
         const svgText = serializeSvgToString(svgTag, true /* shouldInjectFonts */);
         this._svgImageLoaded = false;
 
+        // If there is another load already in progress, replace the old onload to effectively cancel the old load
         this._svgImage.onload = () => {
             const {x, y, width, height} = svgTag.viewBox.baseVal;
             this._size[0] = width;
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -29,7 +29,10 @@ class SVGSkin extends Skin {
         this._renderer = renderer;
 
         /** @type {HTMLImageElement} */
-        this._svgImage = null;
+        this._svgImage = document.createElement('img');
+
+        /** @type {boolean} */
+        this._svgImageLoaded = false;
 
         /** @type {Array<number>} */
         this._size = [0, 0];
@@ -165,7 +168,7 @@ class SVGSkin extends Skin {
         // Can't use bitwise stuff here because we need to handle negative exponents
         const mipScale = Math.pow(2, mipLevel - INDEX_OFFSET);
 
-        if (this._svgImage.complete && !this._scaledMIPs[mipLevel]) {
+        if (this._svgImageLoaded && !this._scaledMIPs[mipLevel]) {
             this._scaledMIPs[mipLevel] = this.createMIP(mipScale);
         }
 
@@ -191,10 +194,10 @@ class SVGSkin extends Skin {
      */
     setSVG (svgData, rotationCenter) {
         const svgTag = loadSvgString(svgData);
-        this._svgImage = document.createElement('img');
         const svgText = serializeSvgToString(svgTag, true /* shouldInjectFonts */);
+        this._svgImageLoaded = false;
 
-        this._svgImage.addEventListener('load', () => {
+        this._svgImage.onload = () => {
             const {x, y, width, height} = svgTag.viewBox.baseVal;
             this._size[0] = width;
             this._size[1] = height;
@@ -218,8 +221,10 @@ class SVGSkin extends Skin {
             this._rotationCenter[0] = rotationCenter[0] - x;
             this._rotationCenter[1] = rotationCenter[1] - y;
 
+            this._svgImageLoaded = true;
+
             this.emit(Skin.Events.WasAltered);
-        });
+        };
 
         this._svgImage.src = `data:image/svg+xml;utf8,${encodeURIComponent(svgText)}`;
     }
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -1,7 +1,7 @@
 const twgl = require('twgl.js');
 
 const Skin = require('./Skin');
-const SvgRenderer = require('scratch-svg-renderer').SVGRenderer;
+const {loadSvgString, serializeSvgToString} = require('scratch-svg-renderer');
 const ShaderManager = require('./ShaderManager');
 
 const MAX_TEXTURE_DIMENSION = 2048;
@@ -28,8 +28,17 @@ class SVGSkin extends Skin {
         /** @type {RenderWebGL} */
         this._renderer = renderer;
 
-        /** @type {SvgRenderer} */
-        this._svgRenderer = new SvgRenderer();
+        /** @type {HTMLImageElement} */
+        this._svgImage = null;
+
+        /** @type {Array<number>} */
+        this._size = [0, 0];
+
+        /** @type {HTMLCanvasElement} */
+        this._canvas = document.createElement('canvas');
+
+        /** @type {CanvasRenderingContext2D} */
+        this._context = this._canvas.getContext('2d');
 
         /** @type {Array<WebGLTexture>} */
         this._scaledMIPs = [];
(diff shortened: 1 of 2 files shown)


src/SVGSkin.js
```diff
@@ -94,18 +103,26 @@ class SVGSkin extends Skin {
      * @return {SVGMIP} An object that handles creating and updating SVG textures.
      */
     createMIP (scale) {
-        this._svgRenderer.draw(scale);
+        const {width, height} = this._svgMeasurements;
+        this._canvas.width = width * scale;
+        this._canvas.height = height * scale;
+        // Even if the canvas at the current scale has a nonzero size, the image's dimensions are floored pre-scaling.
+        // e.g. if an image has a width of 0.4 and is being rendered at 3x scale, the canvas will have a width of 1, but
+        // the image's width will be rounded down to 0 on some browsers (Firefox) prior to being drawn at that scale.
```
adroitwhiz: note to self: move this comment in the middle of the condition so it's clear it refers specifically to the naturalWidth/naturalHeight check

src/SVGSkin.js
```diff
@@ -28,8 +28,17 @@ class SVGSkin extends Skin {
         /** @type {RenderWebGL} */
         this._renderer = renderer;
 
-        /** @type {SvgRenderer} */
-        this._svgRenderer = new SvgRenderer();
+        /** @type {HTMLImageElement} */
+        this._svgImage = null;
+
+        /** @type {Object} */
+        this._svgMeasurements = {x: 0, y: 0, width: 0, height: 0};
```
adroitwhiz: We don't need x and y right now but will need them for #550

src/SVGSkin.js
```diff
@@ -147,7 +165,7 @@ class SVGSkin extends Skin {
         // Can't use bitwise stuff here because we need to handle negative exponents
         const mipScale = Math.pow(2, mipLevel - INDEX_OFFSET);
 
-        if (this._svgRenderer.loaded && !this._scaledMIPs[mipLevel]) {
+        if (this._svgImage.complete && !this._scaledMIPs[mipLevel]) {
```
cwillisf: It's not clear to me that this is guaranteed to never become true before the `load` event handler is called. I think to be safe we should set a flag in the `load` event handler and check that here, similar to how the `loaded` flag worked before.

cwillisf (changes requested):
Thanks for doing this, and sorry it took us so long to review it!
There's one small change I've requested just to make sure it's safe but overall it looks pretty good

src/SVGSkin.js
adroitwhiz: I've done so now. This required some additional supporting changes to avoid a race condition in case `setSVG` is called a second time before the original image is finished loading.

cwillisf (approved):
👍

PR #776, opened 2021-02-11
Add unit tests for color conversion functions

### LINKED ISSUE #678: RGB/HSV color conversion functions should be covered by unit tests
### Expected Behavior

The `rgbToHsv` and `hsvToRgb` color conversion functions in `EffectTransform.js` should be included in unit tests.

### Actual Behavior

Currently those functions are not tested. They're also not exported, which will be necessary to add tests.

adroitwhiz:
### Resolves

Resolves #678

### Proposed Changes

This PR moves `rgbToHsv` and `hsvToRgb` to their own file, and adds unit tests for them. The edge-case-finding part of my brain isn't firing on all cylinders today so feel free to suggest more useful tests.

### Reason for Changes

More tests = more better

### Test Coverage

Yes

### DIFF
--- a/test/unit/ColorConversionTests.js
+++ b/test/unit/ColorConversionTests.js
@@ -1,0 +1,47 @@
+const {test, Test} = require('tap');
+
+const {rgbToHsv, hsvToRgb} = require('../../src/util/color-conversions');
+
+Test.prototype.addAssert('colorsAlmostEqual', 2, function (found, wanted, message, extra) {
+    /* eslint-disable no-invalid-this */
+    message += `: found ${JSON.stringify(Array.from(found))}, wanted ${JSON.stringify(Array.from(wanted))}`;
+
+    // should always return another assert call, or
+    // this.pass(message) or this.fail(message, extra)
+    if (found.length !== wanted.length) {
+        return this.fail(message, extra);
+    }
+
+    for (let i = 0; i < found.length; i++) {
+        if (Math.abs(found[i] - wanted[i]) > 1e-3) {
+            return this.fail(message, extra);
+        }
+    }
+
+    return this.pass(message);
+    /* eslint-enable no-invalid-this */
+});
+
+test('RGB to HSV', t => {
+    const dst = [0, 0, 0];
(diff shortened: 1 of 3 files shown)


test/unit/ColorConversionTests.js
```diff
@@ -0,0 +1,47 @@
...
+    message += `: found ${JSON.stringify(Array.from(found))}, wanted ${JSON.stringify(Array.from(wanted))}`;
+
+    // should always return another assert call, or
+    // this.pass(message) or this.fail(message, extra)
+    if (found.length !== wanted.length) {
+        return this.fail(message, extra);
+    }
+
+    for (let i = 0; i < found.length; i++) {
+        if (Math.abs(found[i] - wanted[i]) > 1e-3) {
```
fsih, line 16: What do you think of 1/256 or 1/360 for epsilon here?

test/unit/ColorConversionTests.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    return this.pass(message);
+    /* eslint-enable no-invalid-this */
+});
+
+test('RGB to HSV', t => {
+    const dst = [0, 0, 0];
+    t.colorsAlmostEqual(rgbToHsv([255, 255, 255], dst), [0, 0, 1], 'white');
+    t.colorsAlmostEqual(rgbToHsv([0, 0, 0], dst), [0, 0, 0], 'black');
```
fsih, line 29: Do we care about rgb(0,0,0) should equal hsv(0,1,0)?

test/unit/ColorConversionTests.js
cwillisf: It'd be nice to use a tolerance that means something, rather than just an arbitrary `> 1e-3`. Maybe `>= 0.5 / 360`, so it'll reject any error that would cause variance in a hue value after rounding?

test/unit/ColorConversionTests.js
fsih: I guess not since that's just a property of the test equals function

fsih (approved):
I think this is good to merge! Just going to leave it open for a bit for the discussion on the tolerance number

fsih (approved):
LG, thanks for tests!

PR #912, opened 2021-06-29
Update svg renderer dependency

fsih:
Update svg renderer to latest, and peer depend on svg render fonts


cwillisf (approved):
👍
Thanks for figuring out the Travis thing, too!

PR #946, opened 2021-07-28
Fix SVG costume dimensions appearing as 0x0 in GUI

### LINKED ISSUE #945: scratch-render currently fails scratch-gui tests
### Actual Behavior

Recent changes in scratch-render are causing this test to fail in scratch-gui:

```
 PASS  test/integration/blocks.test.js (37.63s)
 FAIL  test/integration/costumes.test.js (62.828s)
  ● Working with costumes › Adding an svg from file

    TimeoutError: findByXpath timed out for path: //body//*[@id='react-tabs-3']//*[contains(text(), '100 x 100')]
    Wait timed out after 20007ms
```

See also LLK/scratch-gui#7406

Running the test non-headless, it looks like the costume tab is reporting 0x0 for the size of the costume, where it should be 100x100. Interacting with the paint editor causes the costume size to update, so it's probably just an order-of-operations problem where the size is reported out before it gets properly initialized/updated.

It seems likely that this was introduced with #745 but it could have been anything merged in June 2021.

### Steps to Reproduce

1. Check out `scratch-gui` and update its `scratch-render` dependency to current `scratch-render/develop`
   - Or, check out `scratch-gui` branch `renovate/scratch-render-0.x`
2. Run integration tests

adroitwhiz:
### Resolves

Resolves #945

### Proposed Changes

This PR moves the setting of `SVGSkin`s' `size` outside the callback which fires after the SVG image is loaded.

This isn't an ideal solution, as the skin is technically in an invalid state while the image is still loading--it has the `size` of the new image, but the texture of the old one, and we're relying on not emitting the `WasAltered` event until the new image has loaded, as if the event were to fire before the new image was loaded, the skin's old texture would appear stretched to the new image's dimensions.

Ideally, `setSVG` and `createSVGSkin` would return `Promise`s, and the VM would wait for them before setting the costume's size based on the skin's size, but that would be a breaking change since `createSVGSkin` currently returns the skin ID.

### Reason for Changes

[The VM reads back the skin's size immediately after calling `createSVGSkin`](https://github.com/LLK/scratch-vm/blob/977dfeec383884b33dca5fa53c701e480495668e/src/import/load-costume.js#L26-L27), but the "SVG loaded" callback fires asynchronously, after the VM has already read the skin's size.

### Test Coverage

Since this has happened [before](https://github.com/LLK/scratch-render/pull/497), I've added integration tests that ensure that all skins' dimensions are set synchronously.

### DIFF
--- a/src/SVGSkin.js
+++ b/src/SVGSkin.js
@@ -196,12 +196,17 @@ class SVGSkin extends Skin {
         const svgText = serializeSvgToString(svgTag, true /* shouldInjectFonts */);
         this._svgImageLoaded = false;
 
+        const {x, y, width, height} = svgTag.viewBox.baseVal;
+        // While we're setting the size before the image is loaded, this doesn't cause the skin to appear with the wrong
+        // size for a few frames while the new image is loading, because we don't emit the `WasAltered` event, telling
+        // drawables using this skin to update, until the image is loaded.
+        // We need to do this because the VM reads the skin's `size` directly after calling `setSVG`.
+        // TODO: return a Promise so that the VM can read the skin's `size` after the image is loaded.
+        this._size[0] = width;
+        this._size[1] = height;
+
         // If there is another load already in progress, replace the old onload to effectively cancel the old load
         this._svgImage.onload = () => {
-            const {x, y, width, height} = svgTag.viewBox.baseVal;
-            this._size[0] = width;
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
This is great - thank you! I especially appreciate the new test.

PR #999, opened 2022-10-24
[IBE-566] Migrate to circleci

meyerhot95:
### Resolves

https://scratchfoundation.atlassian.net/browse/IBE-566

### Proposed Changes

Move from travis to circle while retaining maximal continuity with old config.

### Reason for Changes

Deprecating travis

### Test Coverage

Manually tested


.circleci/config.yml
```diff
@@ -0,0 +1,105 @@
...
+    steps:
+      - run:
+          name: Setup Deploy
+          command: |
+            echo "export NPM_TAG=latest" >> $BASH_ENV
+            echo "export NODE_ENV=production" >> $BASH_ENV
+            echo "export RELEASE_TIMESTAMP=$(date +'%Y%m%d%H%M%S')" >> $BASH_ENV
+            npm run build
+            sudo chmod +x .circleci/before_deploy.sh
+            .circleci/before_deploy.sh
```
cwillisf: Since this is the only place it's used, it seems like it would make sense to just move the `before_deploy.sh` contents into this script. Unless there's a reason not to do that...?

If it does need to be in a different file, I'd recommend doing a `chmod +x before_deploy.sh` and then `git commit` that, so you don't need the `chmod` step here.

.circleci/config.yml
```diff
@@ -0,0 +1,105 @@
...
+            - ~/.cache/ms-playwright
+
+  setup:
+    steps:
+      - run:
+          name: Install dependencies
+          command: |
+            npm --production=false install
+            npm --production=false update
+            npx playwright install
```
cwillisf, line 31: Could you explain what the `npx playwright install` does here? I don't see it in the old Travis script.

.circleci/config.yml
```diff
@@ -0,0 +1,105 @@
...
+            - node_modules
+            - ~/.cache/ms-playwright
+
+  setup:
+    steps:
+      - run:
+          name: Install dependencies
+          command: |
+            npm --production=false install
+            npm --production=false update
```
cwillisf, line 30: Oof... this `update` is a little embarrassing in retrospect. I'll clean it up when I do the rest of the modernizing, though :)

cwillisf (changes requested):
Looks good overall, just a couple details to discuss...

.circleci/config.yml
meyerhot95: I believe it installs playwright for testing. I added this because circle was complaining and telling me to do so.

.circleci/config.yml
meyerhot95: Moved script into main file

cwillisf (approved):

PR #1265, opened 2024-06-13
[UEPR-17] Migrate webpack to version 5

MiroslavDionisiev:
### Jira Ticket
[UEPR-17](https://scratchfoundation.atlassian.net/browse/UEPR-17)

### Proposed Changes

Updated webpack to version 5 and other related dependencies. Used `scratch-webpack-configuration` to standardize the format of the webpack configuration between different application. Changed `html` files used by integration tests to be generated by the build. 

### Test Coverage

- Build works, tests continue to pass
- Linked with scratch-vm, tests continue to pass
- Linked with scratch-gui, tests continue to pass

[UEPR-17]: https://scratchfoundation.atlassian.net/browse/UEPR-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cwillisf (commented):
These changes look reasonable. You mentioned that some `scratch-gui` tests fail... do you know why? If it's just because `scratch-gui` needs adjustments to accept these changes, I think it's OK to add a major version bump here and move forward. Maybe it needs a major version bump anyway, for example due to webpack 5's polyfill changes?

MiroslavDionisiev:
@cwillisf, what I meant was integration tests fail locally with and without the changes in `scratch-render`. The error is `ChromeDriver could not be found on the current PATH`. The issue may be in the local setup because I see the check is green on the develop branch on github.

cwillisf:
Ah! This might just be a matter of installing Chrome, Chromium, or a derivative, then. In a pinch, you can `npm i -g chromedriver` but I would treat that as a last resort. The `chromedriver` NPM package is a JS wrapper for the real (binary) `chromedriver`, and the NPM package includes some logic to download and install a `chromedriver` binary. The usual way to get the `chromedriver` binary is to install a Chrome-family browser. Note that the version of the `chromedriver` binary must match the browser version.

As you might have noticed, our tests currently require a Chrome-family browser. Ideally they'd work on other browsers as well, but I'd rather not mix that effort into this one.

